PR #7, opened 2021-02-10
Data reader

bryantChhun:
## description
Data reader that is capable of pulling ome-tiff image stacks, micro-manager metadata, and recast as numpy or zarr.

It currently does not support image sequences (single page tiffs).  These should be for future PRs because the sequence file name formats change depending on the micro-manager version

## method
This reader searches a folder for an ome-tiff master file.  The search reads only omexml metadata and scans for the element tag "BinaryOnly", which indicates a non-master file.  Additionally, micro-manager metadata tags are pulled:
- height (int)
- width (int)
- frames (timepoints)
- slices (int)
- channels (int)
- chnames (channel names)
- stage positions (json group that has detailed stage position info)

I have confirmed that the channel order, as reported in the micro-manager metadata, matches the index order of the array.  This should become an automatic test in the future.

The data is read only as zarr, then recast as numpy array if needed.

@mattersoflight @lihaoyeh @camFoltz


waveorder/io/reader.py
```diff
@@ -0,0 +1,158 @@
...
+        self._set_mm_meta()
+
+    def _set_mm_meta(self):
+        with TiffFile(self.master_ome_tiff) as tif:
+            self.mm_meta = tif.micromanager_metadata
+
+            if self.mm_meta['Summary']['Positions'] > 1:
+                self.stage_positions = []
+                for p in range(self.mm_meta['Summary']['Positions']):
+                    print(f"appending {self.mm_meta['Summary']['StagePositions'][p]}")
```
camFoltz, line 42: I recommend changing this line to 
```suggestion
                    print(f"appending {self.mm_meta['Summary']['StagePositions'][p]['Label']}")
```
This will give the position name instead of listing the long dictionary of elements which can clog up the print statement.

waveorder/io/reader.py
```diff
@@ -0,0 +1,158 @@
...
+                    self.stage_positions.append(self.mm_meta['Summary']['StagePositions'][p])
```
camFoltz, line 43: We may want to do this in a smarter way.  Currently this will append the following dictionary for every position that would look like this: `{'Label': 'C1_p1', 'DefaultXYStage': 'XYStage', 'DefaultZStage': 'ZStage', 'GridRow': 0, 'GridCol': 0, 'DevicePositions': [{'Device': 'FocusDrive', 'Position_um': [6895.653104999999]}, {'Device': 'XYStage', 'Position_um': [27420.3, 6347.5]}, {'Device': 'ZStage', 'Position_um': [22.700000000000003]}]}`

What we can do here is a bit of filtering to clean up this dictionary and make it a bit easier to view.

Some Suggestions could be to make it so that the dictionary elements looks as such: 
`{'Label': 'C1_p1', 'GridRow': 0, 'GridCol': 0, 'DevicePositions': {'FocusDrive': [6895.653104999999]}, 'XYStage': [27420.3, 6347.5], 'Device': 'ZStage', 'Position_um': [22.700000000000003]}}`

waveorder/io/reader.py
```diff
@@ -0,0 +1,158 @@
...
+        self.height = 0
+        self.width = 0
+        self.frames = 0
+        self.slices = 0
+        self.channels = 0
+        self.chnames = []
+
+        self.master_ome_tiff = self._get_master_ome_tiff(folder)
+        print(self.master_ome_tiff)
+        self._set_mm_meta()
```
camFoltz, line 33: Add flag to extract data upon the initiation of the class? I found that this extra step can sometimes be a little annoying if I want to see all of the positions in the class without loading an array.  Default can be false.

```suggestion
    def __init__(self, folder, extract_data=False):
        """
        loads ome-tiff files in folder into zarr or numpy arrays
        Strategy:
            1. read all files in folder
            2. search for the file whose omexml does not contain the element tag "BinaryOnly"-> this is the master ome-tiff
            3. each series represents a micro-manager stage position.  upon call for data, assign this data to a dict

        :param folder: folder containing all ome-tiff files
        :param reader: tifffile or aicsimageio
        """

        self._positions = {}
        self.mm_meta = None
        self.stage_positions = 0
        self.height = 0
        self.width = 0
        self.frames = 0
        self.slices = 0
        self.channels = 0
        self.chnames = []

        self.master_ome_tiff = self._get_master_ome_tiff(folder)
        print(self.master_ome_tiff)
        self._set_mm_meta()
        if extract_data:
            self._extract_data(self.master_ome_tiff)
        
```

waveorder/io/reader.py
camFoltz: It may be better to get rid of these appending statements altogether

waveorder/io/reader.py
```diff
@@ -0,0 +1,158 @@
...
+            self.mm_meta = tif.micromanager_metadata
+
+            if self.mm_meta['Summary']['Positions'] > 1:
+                self.stage_positions = []
+                for p in range(self.mm_meta['Summary']['Positions']):
+                    print(f"appending {self.mm_meta['Summary']['StagePositions'][p]}")
+                    self.stage_positions.append(self.mm_meta['Summary']['StagePositions'][p])
+
+            for ch in self.mm_meta['Summary']['ChNames']:
+                print(f"appending {ch}")
```
camFoltz, line 46: Suggest removing this statement

```suggestion

```

waveorder/io/reader.py
```diff
@@ -0,0 +1,158 @@
...
+        if len(self._positions) == 0:
+            self._extract_data(self.master_ome_tiff)
+        return np.array(self._positions[position])
+
+    def get_master_ome(self):
+        return self.master_ome_tiff
+
+    def get_num_positions(self):
+        return len(self._positions)
+
```
lihaoyeh, line 137: `get_num_positions` will return 0 before using `get_array` or `get_zarr`. Is this something expected?

waveorder/io/reader.py
lihaoyeh: I agree with this suggestion. Printing detailed metadata upfront seems a bit confusing.

waveorder/io/reader.py
```diff
@@ -0,0 +1,158 @@
...
+        extract all series from ome-tiff and place into dict of (pos: zarr)
+        :param master_ome: full path to master OME-tiff
+        :return:
+        """
+        with TiffFile(master_ome) as tif:
+            for idx, tiffpageseries in enumerate(tif.series):
+                self._positions[idx] = zarr.open(tiffpageseries.aszarr(), mode='r')
+
+    def _get_master_ome_tiff(self, folder_):
+        """
```
lihaoyeh, line 66: Using the ome.tiff datasets you provided, I got a warning like:
image

What does this mean and do we need to pay attention to it?

waveorder/io/reader.py
```diff
@@ -0,0 +1,158 @@
+import numpy as np
+import os
+import zarr
+from tifffile import TiffFile
+
+
+class MicromanagerReader:
+
+    def __init__(self, folder):
```
ieivanov, line 9: It will be best to turn off all print statements by default, and optionally add a verbose mode

waveorder/io/reader.py
```diff
@@ -0,0 +1,158 @@
...
+        """
+        extract all series from ome-tiff and place into dict of (pos: zarr)
+        :param master_ome: full path to master OME-tiff
+        :return:
+        """
+        with TiffFile(master_ome) as tif:
+            for idx, tiffpageseries in enumerate(tif.series):
+                self._positions[idx] = zarr.open(tiffpageseries.aszarr(), mode='r')
+
+    def _get_master_ome_tiff(self, folder_):
```
ieivanov, line 65: On dataset `/gpfs/CompMicro/rawdata/falcon/Ivan/20210128 HEK CAAX SiRActin/FOV1_1` this search incorrectly identifies `FOV1_1_MMStack_Default_23.ome.tif` as the master ome.tif file. I also get `invalid MicroManager display header` and `invalid MicroManager comments header` warnings. The code then proceeds to print "appending ... channels"

waveorder/io/reader.py
```diff
@@ -0,0 +1,158 @@
...
+        :param reader: tifffile or aicsimageio
+        """
+
+        self._positions = {}
+        self.mm_meta = None
+        self.stage_positions = 0
+        self.height = 0
+        self.width = 0
+        self.frames = 0
+        self.slices = 0
```
ieivanov, line 27: This object should have a `shape` properties which returns the data dimensions in known order, say `TCZYX`. We can then map Micromanager ```self.frames = self.shape[0], self.channels = self.shape[1]```, etc. for convenience.

If our plan is to eventually switch to `aicsimageio`, then we should try to emulate their API as much as possible.

waveorder/io/reader.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def _extract_data(self, master_ome):
+        """
+        extract all series from ome-tiff and place into dict of (pos: zarr)
+        :param master_ome: full path to master OME-tiff
+        :return:
+        """
+        with TiffFile(master_ome) as tif:
+            for idx, tiffpageseries in enumerate(tif.series):
+                self._positions[idx] = zarr.open(tiffpageseries.aszarr(), mode='r')
```
ieivanov, line 63: This will create zarr stores for all series in the given ome.tif. This can create unnecessary overhead if you've only asked for one position. On the IBM clusters currently creating a zarr store with `store = tifffile.imread(path, aszarr=True)` takes an inordinate amount of time, tho I haven't tried the syntax you're using here.

waveorder/io/reader.py
```diff
@@ -0,0 +1,158 @@
...
+        self.frames = 0
+        self.slices = 0
+        self.channels = 0
+        self.chnames = []
+
+        self.master_ome_tiff = self._get_master_ome_tiff(folder)
+        print(self.master_ome_tiff)
+        self._set_mm_meta()
+
+    def _set_mm_meta(self):
```
ieivanov, line 35: Micromanager writes Summary metadata and per-image metadata. It would be useful to be able to retrieve metadata for a specific image at given TCZYX coordinates. I tried doing something like that with this function, it may need some debugging.

https://github.com/czbiohub/aicsimageio/blob/30fa17a1fa0b1e9351dea610972e438c87185a4a/aicsimageio/readers/micromanager_reader.py#L114-L132

ieivanov (changes requested):

waveorder/io/reader.py
bryantChhun: From what I can see, you're asking to reformat the `DevicePositions` from 
- a list of dictionaries, where each list element is a device dictionary, and each device dictionary contains `Device` and `Position_um` keys
to
- a dictionary, where each key is the device name (value of `Device` from above), and each value is the list of int/float (the `Position_um` from above)

In this case I might correct your suggestion to look like:
`{'Label': 'C1_p1', 'GridRow': 0, 'GridCol': 0, 'DevicePositions': {'FocusDrive': [6895.653104999999], 'XYStage': [27420.3, 6347.5], 'ZStage': [22.700000000000003] } }`

Then the access pattern changes from:
```
>>> self.mm_meta['Summary']['StagePositions'][p]['DevicePositions'][0]['Device']
'FocusDrive' 
>>> self.mm_meta['Summary']['StagePositions'][p]['DevicePositions'][0]['Position_um']
[6895.653104999999]
```
to
```
>>> self.mm_meta['Summary']['StagePositions'][p]['DevicePositions']['FocusDrive']
[6895.653104999999]
>>> self.mm_meta['Summary']['StagePositions'][p]['DevicePositions']['XYStage']
[27420.3, 6347.5]
```

One complication i see is that you have to know the device name ( `FocusDrive` or `XYStage` ) which can vary depending on the mm-config.  It's harder to programmatically search for a key that may or may not exist (which would return an error if does not exist).

It's easy enough to fix this dictionary structure.  There are also other things we can do -- write a method to parse values out and assign them to class attributes?

waveorder/io/reader.py
bryantChhun: agreed, will implement this.

waveorder/io/reader.py
camFoltz: You wouldn't have to search for the matching key, since it is already in the metadata. For every position, you can first store the names of the stage devices, and then loop through those names to find their values

```
devices = []
for i in range(len(self.mm_meta['Summary']['StagePositions'][p]['DevicePositions'][0]):
    devices.append(self.mm_meta['Summary']['StagePositions'][p]['DevicePositions'][i]['Device'])

stage_positions = []
for dev in devices:
    stage_positions.append({dev: float(self.mm_meta['Summary']['StagePositions'][p]['DevicePositions'][dev]})
```

bryantChhun: > ```
> devices = []
> for i in range(len(self.mm_meta['Summary']['StagePositions'][p]['DevicePositions'][0]):
>     devices.append(self.mm_meta['Summary']['StagePositions'][p]['DevicePositions'][i]['Device'])
> 
> stage_positions = []
> for dev in devices:
>     stage_positions.append({dev: float(self.mm_meta['Summary']['StagePositions'][p]['DevicePositions'][dev]})
> ```

I see.  So the ask is to have stage_positions list contain only the device info, not `Label`, `Grid_row`, or `Grid_col`:
Is that right?

Before:
```
>>> stage_positions[0]
{'Label': 'C1_p1', 
'DefaultXYStage': 'XYStage', 
'DefaultZStage': 'ZStage', 
'GridRow': 0, 
'GridCol': 0, 
'DevicePositions': [{'Device': 'FocusDrive', 'Position_um': [6895.653104999999]}, 
                    {'Device': 'XYStage', 'Position_um': [27420.3, 6347.5]}, 
                    {'Device': 'ZStage', 'Position_um': [22.700000000000003]}
                    ]
}
```
after:
```
>>> stage_positions[0]
 {'FocusDrive': [6895.653104999999]},
  'XYStage': [27420.3, 6347.5], 
  'ZStage': [22.700000000000003]}

File: waveorder/io/reader.py (hunk 5 above)
Comment (bryantChhun) [reply waveorder/io/reader.py:137]: I'm not sure what I intended with this method.  It returns the number of scenes in the multipage-tiff, but as you note, only if you've already initialized with `get_array` or `get_zarr`.  Maybe it's not worth existing?

File: waveorder/io/reader.py (hunk 6 above)
Comment (bryantChhun) [reply waveorder/io/reader.py:66]: Sadly this is a result of tifffile library.  I'd have to do some digging to find out what is invalid about the micro-manager JSON.  There must be some formatting standard that micro-manager does not adhere to.

File: waveorder/io/reader.py (hunk 8 above)
Comment (bryantChhun) [reply waveorder/io/reader.py:65]: I will explore this.  I believe in older versions of micro-manager, every file associated with the ome-tiff had the master scene info, so it didn't matter which one the class selects.  For time series, this might be true too (?).

The warnings are probably output from tifffile.  It seems to care about json formatting.  If this is the case, the proper place to fix it is within micro-manager.

File: waveorder/io/reader.py (hunk 6 above)
Comment (mattersoflight) [reply waveorder/io/reader.py:66]: I recently discussed OME-TIFFs and OME-zarr's with Nico. He pointed out that the data formats that work the best for image acquisition (where you want to write images as they come) and image analysis (where you want to structure images such that they are easy to parse) is different. For this reason, Micro-Manager OME-TIFFs are not fully compatible with OME-TIFF standard. Does this error affect the reading of planes? If not, it is safe to ignore.

File: waveorder/io/reader.py (hunk 10 above)
Comment (bryantChhun) [reply waveorder/io/reader.py:63]: From what I understand, the major overhead occurs with calls to `tif.series`.  In Tifffile's code, it appears this call parses through the entire ome structure to get the page locations of each scene, then returns the list of pages (`tiffpageseries` above).

I could try assigning the ZarrTiffStore to self._positions[idx] rather than the zarr array.  But I found the memory footprint was low using the zarr array, and so assumed there was lazy loading anyways.

File: waveorder/io/reader.py (hunk 6 above)
Comment (bryantChhun) [reply waveorder/io/reader.py:66]: Interesting point and definitely worth keeping in mind as we work with Allen Inst.
For now i'm implementing a logger that is the same as what Aicsimageio folks use.  You can set the reporting level to `ERROR` or `CRITICAL` which will suppress the JSON and other warnings even from tifffile library.

File: waveorder/io/reader.py (hunk 10 above)
Comment (bryantChhun) [reply waveorder/io/reader.py:63]: Ok after some testing, I can't find any obvious way to interface with `tifffile` library to speed this up.  Any call to `tif.series` results in a calls to `_load` and then `_seek`, which scans all pages to identify their offset locations.  I'm not sure there's an easy solution here, as even aicsimageio uses tifffile to access data this way.

File: waveorder/io/reader.py (hunk 11 above)
Comment (bryantChhun) [reply waveorder/io/reader.py:35]: This could be a cool feature, but perhaps it deserves another PR.  Your link references the current state of aicsimageio, which uses [its own OME class](https://github.com/czbiohub/aicsimageio/blob/30fa17a1fa0b1e9351dea610972e438c87185a4a/aicsimageio/vendor/omexml.py#L302) to parse the files.  The future of aicsimageio uses [`ome_types` library ](https://github.com/tlambert03/ome-types).

The worry is that, ome_types threw a bunch of errors when loading micro-manager data when using aicsimageio.  If we use it here, it might also give us problems.  Will need to dive deeper.

File: waveorder/io/reader.py (hunk 7 above)
Comment (bryantChhun) [reply waveorder/io/reader.py]: verbose mode now implemented

File: waveorder/io/reader.py (hunk 9 above)
Comment (bryantChhun) [reply waveorder/io/reader.py:54]: This is reasonable but I want to stop short of implementing specific features that are well-enabled by aicsimageio.  (for example, per-image metadata parsing or other xml-specific operations).  Maybe this requires a rethink of the short term and long term goals of this reader.

File: waveorder/io/reader.py (hunk 8 above)
Comment (ieivanov) [reply waveorder/io/reader.py:107]: >I believe in older versions of micro-manager, every file associated with the ome-tiff had the master scene info, so it didn't matter which one this reader selects. For time series, this might be true too (?).

I believe the version I'm using is from ~June last year. I can easily switch to a newer version, I'm not particularly tied to it.

> It seems to care about json formatting. If this is the case, the proper place to fix it is within micro-manager.

I recently learned how to suppress certain warnings: https://docs.python.org/3/library/warnings.html. Either that, or we fix the issue in MM, as you said.

Comment (bryantChhun) [reply waveorder/io/reader.py:107]: >I recently learned how to suppress certain warnings: https://docs.python.org/3/library/warnings.html. Either that, or we fix the issue in MM, as you said.
The recent pushes to this reader have a flag to set logging level.  By default it will be set to logging.ERROR which suppresses warnings.

File: waveorder/io/reader.py (hunk 10 above)
Comment (ieivanov) [reply waveorder/io/reader.py:105]: Ah, I see, I didn't know that about `tif.series`. My suggestion here was to simply remove the for loop over tif.series and create a zarr store only for the series we're currently working with.

On that note, I'm not sure we should be treating the position dimension in the same way as all other dimensions. As you know, data in different series can have different dimensions, tho on the other hand MM acquisition doesn't really allow you to write data that way. Using `pycromanager` it is possible to write data of different dimensions at separate positions, but then again here we're developing a reader that's pretty tailored to MicroManager.

I think it'll be a good to separate the position dimensions from the rest. We can have methods like `set_series` to specify which position you want data for. We can also have `get_number_of_series` which will return the number of positions. That seems to be the approach that `aicsimageio` is taking in API v4.0 (https://github.com/AllenCellModeling/aicsimageio/tree/main). What do you think?

File: waveorder/io/reader.py (hunk 11 above)
Comment (ieivanov) [reply waveorder/io/reader.py:65]: I see, I haven't kept up to date with the way metadata reading has changed. We can certainly leave then for a future PR

File: waveorder/io/reader.py (hunk 10 above)
Comment (bryantChhun) [reply waveorder/io/reader.py:105]: >I think it'll be a good to separate the position dimensions from the rest. We can have methods like set_series to specify which position you want data for. 

`self._positions` is a dictionary whose keys are scene/position indicies.  So this would enable each key's value to have its own dimensions.

>We can also have get_number_of_series which will return the number of positions. That seems to be the approach that aicsimageio is taking in API v4.0 (https://github.com/AllenCellModeling/aicsimageio/tree/main). What do you think?

Ah!  I just removed this method but can put it back.  Li-Hao had a question about its default value when the data is not multiposition.

**Comment by lihaoyeh** [comment]
Hi all,

I saw Bryant resolved many of the comments in the past few weeks. Do all of you on board with the current PR? If so, I can approve and Bryant can proceed to merge it. How's that sound?

**Comment by camFoltz** [comment]
I am on board with merging this PR.  I believe those changes suggested by Ivan need to be resolved for this to move forward.

**Review by lihaoyeh — DISMISSED** [review]

File: waveorder/io/reader.py (hunk 8 above)
Comment (ieivanov) [reply waveorder/io/reader.py:107]: On dataset `/gpfs/CompMicro/rawdata/falcon/Ivan/20210128 HEK CAAX SiRActin/FOV1_1` the `_get_master_ome_tiff` still incorrectly identifies `FOV1_1_MMStack_Default_23.ome.tif` as the master-ome file and the reader doesn't work. @bryantChhun can you take a look at this again?

Also, if I pass `/gpfs/CompMicro/rawdata/falcon/Ivan/20210108 HEK_CAAX_SiRActin` as the data path, the reader errors out with:
```
~/Documents/waveorder/waveorder/io/reader.py in _get_master_ome_tiff(self, folder_)
    149                     continue
    150 
--> 151         return os.path.join(folder_, ome_master)
    152 
    153     def get_zarr(self, position):

UnboundLocalError: local variable 'ome_master' referenced before assignment
```
`_get_master_ome_tiff` should not error out if we pass a folder that doesn't contain MM data or if `ome_master` can't be found

File: waveorder/io/reader.py [hunk 12]
```diff
@@ -0,0 +1,208 @@
...
+    def get_array(self, position):
+        """
+        return a numpy array for a given position
+        :param position: int
+            position (aka ome-tiff scene)
+        :return: np.ndarray
+        """
+        if not self._positions:
+            self.extract_data(self.master_ome_tiff)
+        return np.array(self._positions[position])
```
ieivanov, line 173: This can be simplified to
```
return np.array(self.get_zarr(position))
```

waveorder/io/reader.py
```diff
@@ -0,0 +1,208 @@
...
+        :param stage_pos: dictionary containing a single position's device info
+        :return:
+        """
+        out = copy(stage_pos)
+        out.pop('DevicePositions')
+        for dev_pos in stage_pos['DevicePositions']:
+            out.update({dev_pos['Device']: dev_pos['Position_um']})
+        return out
+
+    def extract_data(self, master_ome):
```
ieivanov, line 96: Shouldn't this also be an internal function (i.e. `_extract_data`)? Also, I'm not sure about the name of the function. Right now, it sounds like it extract byte data, while in practice it simply creates zarr stopres for every position. Maybe call it `_create_zarr_store`?

waveorder/io/reader.py
bryantChhun: I'll take a look at this today.  Now knowing that the non-master files should contain a pointer to the master, I could refactor this to be more efficient.

waveorder/io/reader.py
ieivanov: I still don't think we should have a for loop over the series (`for idx, tiffpageseries in enumerate(tif.series)`). This create unnecessary overhead, when in practice we most often analyze one position at a time, and it's uncommon to switch back and forth between reading data from different positions, i.e. we sequentially go through all positions.

I think we should have a `set_position` or `set_series` function which creates a zarr store for the series/position the user wants to analyze. Call to this function should be optional, and if not called (the user calls `get_array` before `set_position`), it should default to series=0. We should also have a matching `get_series` so the user can confirm which series they're currently working with.

In this case one workflow will look like that:
```
data_obj = MicromanagerReader('data_path')
data = data_obj.get_array()
## analyze data
```
This will not create zarr stores for other positions in `data_path`.

Another workflow will look like this:
```
data_obj = MicromanagerReader('data_path')
data_obj.set_position(0)
data = data_obj.get_array() ## or data = data_obj.get_array(position=0)
## analyze data

data_obj.set_position(1)
data = data_obj.get_array()
## analyze data
```

`_create_zarr_store` will contain a section something like that:
```
def _create_zarr_store(self, position):
# self._position is the index of the current position/series
# self.zarr_store contains the zarr store
    if self._position != position:
        self._position = position
        self.zarr_store = zarr.open(tif_series, mode = 'r')
```

bryantChhun: I don't think it's possible to index directly into a position without first looping over the whole series -- this is the `tifffile` inefficiency.  I could be wrong and will look into this more.

But even if we must loop over the series, we can enable selective position calls still.  It'll just mean each position call can take variable amounts of time.

ieivanov: Hm, interesting, I didn't know that. I thought you could create a TiffSeries object, but I see that that's not a thing. Ok,  I know you've looked a lot more into this, we'll go with your recommendation, you're welcome to resolve this issue

bryantChhun: I think your suggestion is a good one though.  It doesn't make sense to index every position when we just want one or two.  Am finishing some other repo dev and will get to this today :-)

waveorder/io/reader.py
bryantChhun: I have restructured the way master-files are found.  It now identifies it directly from the metadata without iteration.

waveorder/io/reader.py
bryantChhun: After thinking about this some, it might be very difficult to speed up calls to `tif.series` Likely we would need to clone and re-implement the `tifffile` library.  Let's leave this for later testing/research/

waveorder/io/reader.py
```diff
@@ -0,0 +1,219 @@
...
+
+        self._positions = {}
+        self.mm_meta = None
+        self.stage_positions = 0
+        self.height = 0
+        self.width = 0
+        self.frames = 0
+        self.slices = 0
+        self.channels = 0
+        self.shape = (self.frames, self.slices, self.channels, self.height, self.width)
```
ieivanov, line 61: self.shape is not updated after creating reader object

waveorder/io/reader.py
```diff
@@ -0,0 +1,219 @@
...
+        self.chnames = []
```
ieivanov, line 62: better to rename this to self.channel_names. It's more pythonic, tho I realize it breaks away from the MM naming. Up to you Bryant

waveorder/io/reader.py
bryantChhun: good catch, fixed

waveorder/io/reader.py
bryantChhun: i renamed it.  I prefer channel_names

waveorder/io/reader.py
bryantChhun: i think they're equivalent.  This way there are fewer class function calls

ieivanov (approved):

PR #9, opened 2021-03-30
Basic Writer Structure

camFoltz:
I am starting a PR for the basic writer structure.  This writer allows you to initialize a blank array of specified dimensions / data-type, and then later place data into the array.  This version excludes hierarchies for now, so you would have to create a new zarr store for every reconstruction channel (phase, retardance, S0, S1, ... , etc.).  There are convenience functions to switch around which zarr store you currently have open, or to add a zarr store to the current save directory.  There is also a builder structure implemented for when we eventually add metadata based on the type of data we are saving.  For now, both builders do the same exact thing, but it may be useful to have this separation in the future when metadata / attributes come into play.  However we are open to ideas on this layout.  Here is code snippet on an example use case:

```
from writer import WaveorderWriter
import numpy as np
import zarr

writer = WaveorderWriter('/Users/cameron.foltz/Desktop/zarr_test', 'stokes')
writer.add_store('func_test_5.zarr')

data_shape = (10, 1, 65, 2048, 2048)
chunk_size = (1, 1, 1, 2048, 2048)
dtype = 'uint16'

data = np.ones(data_shape, dtype=dtype)

writer.init_array(data_shape, chunk_size, dtype)

writer.write(data[0, 0, 0:65, :, :], P=[0], T=[0], Z=[0,65])
```


waveorder/io/writer.py
```diff
@@ -0,0 +1,297 @@
...
+        # if not, create that directory
+        if os.path.isdir(path):
+            self.save_dir = path
+        else:
+            os.mkdir(path)
+            self.save_dir = path
+
+    def add_store(self, name: str):
+        # Add Zarr store to existing save directory,
+        # if store already exists, open that store
```
bryantChhun, line 68: I don't think the writer should read stores at all.  Maybe just check if it exists and asks to overwrite or not.

This raises a question of, what do we do when someone wants to load and modify an existing .zarr store?  This codebase is focused on reading mm tiffs (reader) and writing comp.micro reconstructions to zarr (writer) ... Maybe for post-compute modifications we leave that to the standard zarr library?

waveorder/io/writer.py
```diff
@@ -0,0 +1,297 @@
...
+class WaveorderWriter:
+    """
+    given stokes or physical data, construct a standard hierarchy in zarr for output
+        should conform to the ome-zarr standard as much as possible
+
+    """
+    __builder = None
+    save_dir = None
+    store_path = None
+    store = None
```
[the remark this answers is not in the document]
bryantChhun: One caveat to letting them remain public is that you won't be able to call `self._check_is_dir` without overriding `__setattr__` or something similar.

waveorder/io/writer.py
```diff
@@ -0,0 +1,297 @@
...
+
+    # create a directory store and write the memory store to that
+    def to_disk(self): pass
+
+
+class PhysicalZarr(Builder):
+    __pzarr = None
+    __compressor = None
+
+    def write(self, data, P, T, Z):
```
bryantChhun, line 166: `def write` can also be in the parent class `Builder`

waveorder/io/writer.py
```diff
@@ -0,0 +1,297 @@
...
+        """
+        Write data to specified index of initialized zarr array
+
+        :param data: (nd-array), data to be saved. Must be the shape that matches indices (P, T, Z, Y, X)
+        :param P: (tuple or value), index or index range of the position dimension
+        :param T: (tuple or value), index or index range of the time dimension
+        :param Z: (tuple or value), index or index range of the Z dimension
```
bryantChhun, line 173: your use case earlier:
```
writer.write(data[0, 0, 0:65, :, :], P=[0], T=[0], Z=[0,65])
```
uses `lists` for P, T, Z.  So you will want to include that type in this docstring.

Additionally, calling `len(P)` when P is a "value" (assuming int) throws TypeError -- you'll need an explicit check for this.

bryantChhun (changes requested):
overall I like it.  A few comments.

waveorder/io/writer.py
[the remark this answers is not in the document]
camFoltz: I think having the store be available to the user is nice to quickly check if the data and other metadata was written correctly.

I have made the rest private.

waveorder/io/writer.py
camFoltz: Agreed, check now implemented.

waveorder/io/writer.py
camFoltz: I think that if reconstruction went poorly and you want to go back to a certain dataset to overwrite it, then being able to quickly switch stores is useful.  This function has since been replaced by "create_zarr()".

I do think we need some sort of safeguard to prevent the user from accidentally overwriting the data.  Maybe we can prompt the user to agree before continuing if an overwrite is possible.  There is definitely utility in being able to overwrite the data as you play around with reconstruction parameters.

bryantChhun:
@camFoltz 
can you delete the binaries in your recent push?  the `.pyc` files.  They'll be regenerated but shouldn't be version tracked.
we can also add the `waveorder/io/__pycache__/` to the .gitignore

bryantChhun:
> ```
> from writer import WaveorderWriter
> import numpy as np
> import zarr
> 
> writer = WaveorderWriter('/Users/cameron.foltz/Desktop/zarr_test', 'stokes')
> writer.add_store('func_test_5.zarr')
> 
> data_shape = (10, 1, 65, 2048, 2048)
> chunk_size = (1, 1, 1, 2048, 2048)
> dtype = 'uint16'
> 
> data = np.ones(data_shape, dtype=dtype)
> 
> writer.init_array(data_shape, chunk_size, dtype)
> 
> writer.write(data[0, 0, 0:65, :, :], P=[0], T=[0], Z=[0,65])
> ```

the above example is out of date.  Please update.

mattersoflight:
The output of the writer can be loaded easily with napari and Fiji N5 plugin 🎉. napari and N5 plugin don't recognize the channel dimensions and assign them based on order. Is there an attribute that can be set that at least napari can recognize?
image

waveorder/io/writer.py
```diff
@@ -0,0 +1,478 @@
...
+        If no name is supplied, the default builder name will
+        be used ('stokes_data' or 'physical_data')
+
+        Parameters
+        ----------
+        name:       (string) Optional. Name of the zarr store.
+
+        """
+
+        self.__builder.init_zarr(self.position_dir, name)
```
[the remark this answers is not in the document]
bryantChhun: I implemented this

bryantChhun (approved):
I made some changes that are reflected in the commit history.  Of note, which I would like you to chime in on, I have changed the behavior of `create_zarr` and `set_zarr` to be "create zarr directory store and instantiate builders" and "set an existing directory store and instantiate builders" respectively.

This enables one to separate the use stage:
#### set paths for directory store and create them
```
writer = WaveorderWriter('path_to_dir', 'physical')
writer.set_position(0)
writer.create_zarr()
```
or
#### set paths for an existing directory store and load it
```
writer = WaveorderWriter('path_to_dir', 'physical')
writer.set_position(0)
writer.set_zarr(writer.position_dir)
```

next

#### initialize the array
```
writer.init_array(data_shape=(...), chunk_size ....)
```
if using `set_zarr`, `init_array` will throw an error if the loaded zarr was already initialized.  If it was already initialized, `set_zarr` allows you to write to it immediately.

finally

#### write data to zarr
```
writer.write(data[0, 0, 0:10, :, :], t=[0], c=[0], z=[0,10])
```

bryantChhun (dismissed):

bryantChhun (approved):

PR #10, opened 2021-03-31
expanded data reader

bryantChhun:
## What's in this PR

* singlepage tiff sequence reader
* refactor to accomodate reader expansion

This PR expands on the old data reader by:
1. `MicromanagerReader` is now a higher-level function that calls subclasses specific to the data type
2. Subclasses are now `MicromanagerOmeTiffReader` and `MicromanagerSequenceReader`
3. The new structure allows us to expand the types of readers for different formats, with little change to existing code.  Other types of readers we may need: Pycromanager, zarr, ome-zarr, or waveorder-specific zarr.

The added benefit of this is that eventual integration into `aicsimageio` repo may be easier.

## Is it tested?

I've tested this data on mm2.0-gamma and mm1.4.22 data generated both by demo config and experimentally.  Some of the [demo data is here:](https://drive.google.com/drive/folders/1hbpL1Ei3q86L6oq63QwZEp7c33H-asRb?usp=sharing)

For the sequence reader, I tested on a large set of data (50 timepoints, 4 positions, 3 channels, 5 slices, 2k x 2k) (approx 25 GB data) while monitoring process memory.  The process memory does not appear to explode.


waveorder/io/reader.py
```diff
@@ -202,9 +329,254 @@ def shape(self):
...
+            self.log.error("singlepage tiffs not loaded")
+
+    def _create_stores(self):
+        """
+        extract all singlepage tiffs at each coordinate and place them in a zarr array
+        coordinates are of shape = (pos, time, channel, z)
+        :return:
+        """
+        self.log.info("")
+        z = zarr.zeros(shape=(self.frames,
```
camFoltz, line 418: In the OME Tiff reader, the zarr stores created are of the dimension (C, Z, Y, X).  Here we have created stores of (P, C, Z, Y, X) dimensions, with the P dimensions always equal to 1.  I think that we should remove this dimension to keep it consistent with the other reader and because it is implied when you call a certain position with `self.get_array(pos)`

bryantChhun: good catch.  But to clarify -- the coordinate is (P, T, C, Z, Y, X).  Here, zarr is opening a (T, C, Z, Y, X) and placing it in `self.positions` at P

camFoltz: I see, this may be okay then.  Maybe we should then expand the ometiff reader to always include a time dimension.  Or here we can only include the time dimension when it is present, as I believe is the case for the ometiff reader

ieivanov: For reconstruction, it is very useful if the raw data is in consistent dimensions. For example, if you'd like to do a z projection you can simply average along the third dimensions without having to worry how many channels or time points were in this specific dataset. I'd suggest modifying the reader such that data is always returned in (T, C, Z, Y, X) dimensions order, padding with size=1 dimensions where necessary.

I believe this is also what aicsimageio does. The OME tif metadata contains information on the size of each dimension.

bryantChhun: @ieivanov Ok this should be fixed in now.  The dimensions are now consistently (T, C, Z, Y, X) when reading ome-tiff

camFoltz (approved):
Looks good thus far.  Will merge and we can fix small bugs as they come up.

PR #12, opened 2021-05-04
init github actions

bryantChhun:
following github instructions on initializing for continuous integration
[](https://docs.github.com/en/actions/guides/setting-up-continuous-integration-using-workflow-templates)

Edit:
This creates a workflow file: `.github/workflows/pytests.yml`. This is a template provided by github and modified for our use.  What it accomplishes is primarily:
1. upon any push, run flake8 linter and pytest on ubuntu

### DIFF
--- a/.github/workflows/pytests.yml
+++ b/.github/workflows/pytests.yml
@@ -14,37 +14,37 @@ jobs:
         python-version: [3.7]
 
     steps:
-      - name: Checkout repo
-      - uses: actions/checkout@v2
+    - name: Checkout repo
+      uses: actions/checkout@v2
 
-      - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v1
-        with:
-          python-version: ${{ matrix.python-version }}
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.python-version }}
 
-      - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip
-          pip install flake8
-          pip install pytest
-          pip install pytest-cov
-          pip install -r requirements.txt
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install flake8
+        pip install pytest
+        pip install pytest-cov
+        pip install -r requirements.txt
 
-      - name: Lint with flake8
-        run: |
-          pip install flake8
-          # stop the build if there are Python syntax errors or undefined names
-          flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
-          # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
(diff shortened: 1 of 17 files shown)


camFoltz:
It doesn't seem as though this built?

camFoltz:
Looks like the Flake8 linter doesn't like the 'cp' (cupy) in the reconstruction code, Li-Hao wrote it such that there is not a strict important statement, but rather a specification of the global variables here:

```
if use_gpu:
        globals()['cp'] = __import__("cupy")
        cp.cuda.Device(gpu_id).use()
```

The linter then doesn't recognize cp being defined anywhere in the script

bryantChhun:
> Looks like the Flake8 linter doesn't like the 'cp' (cupy) in the reconstruction code, Li-Hao wrote it such that there is not a strict important statement, but rather a specification of the global variables here:
> 
> ```
> if use_gpu:
>         globals()['cp'] = __import__("cupy")
>         cp.cuda.Device(gpu_id).use()
> ```
> 
> The linter then doesn't recognize cp being defined anywhere in the script

I see.  We should ask Li-Hao about this and if it's important.  Maybe the idea is that, if you are not using GPU, the module never imports cupy?  But can he simply replace `globals()['cp'] = __import__("cupy")` with `import cupy as cp` ?

camFoltz:
Yeah that's right.  I think it is because the version of cupy that you need to install is highly dependent on your GPU, so it's hard to list a single version as a requirement, thus you never import cupy unless you want to use the gpu-- assuming you already have the correct version installed.  I agree that `import cupy as cp` would achieve the same thing here though.  It will be good to ask @lihaoyeh

bryantChhun:
>Yeah that's right. I think it is because the version of cupy that you need to install is highly dependent on your GPU, so it's hard to list a single version as a requirement
if that's the case, we might need to think about configuring pypi for optional dependencies.

lihaoyeh:
> > Yeah that's right. I think it is because the version of cupy that you need to install is highly dependent on your GPU, so it's hard to list a single version as a requirement
> 
> if that's the case, we might need to think about configuring pypi for optional dependencies.


Thanks for raising this question. The reason to have `globals()['cp'] = __import__("cupy")` is to import cupy for global use (access cp from all the methods in the class) when `use_gpu=True`. `import cupy as cp` inside a constructor does not allow access of `cp` from other methods in the class. 

There is an alternative way of making `cp` global inside a constructor:
`global cp`
`import cupy as cp`

I am not sure if this alternative resolves this question.

bryantChhun:
> It looks like there is still a failure in the github actions, but I don't see any linting or any other checks like the previous commit. What exactly changed in the actions?

I changed the workflow config file.  Let me see if there's an error now.

bryantChhun:
ok i needed to remove the linter and code cov.  Linter because we have to think about li-hao's approach to cupy loading.  That solution could be its own PR.  Removed the codecov runner because currently, codecov.io does not have access to this repo (shalin will need to give authorization).

camFoltz (approved):

PR #13, opened 2021-05-06
Pytests reader

bryantChhun:
conftests and pytests for the reader class plus its subclasses.

also contains a bug fix in the reader where metadata is string-parsed to create coordinate-filename maps


tests/reader/test_multipagetiff.py
```diff
@@ -0,0 +1,117 @@
...
+    mmr = MicromanagerOmeTiffReader(rand_folder, extract_data=True)
+    for i in range(mmr.get_num_positions()):
+        z = mmr.get_zarr(i)
+        assert(z.shape == mmr.shape)
+
+
+def test_get_array_mm2gamma(setup_mm2gamma_ome_tiffs):
+    _, _, rand_folder = setup_mm2gamma_ome_tiffs
+    mmr = MicromanagerOmeTiffReader(rand_folder, extract_data=True)
+    for i in range(mmr.get_num_positions()):
```
camFoltz, line 51: For the get_array and get_zarr tests, do we want to assert that one is a zarr object and one is a numpy array?  This test could pass if the get_array returned a zarr array or vice versa

bryantChhun: good idea, will do

bryantChhun: done

camFoltz (approved):
Looks good, all the tests passed on my end

PR #18, opened 2021-05-19
2D QLIPP reconstruction test

lihaoyeh:
This PR adds a test of 2D QLIPP reconstruction. The reconstruction of simulated data is benchmarked against the ground true images. Reconstructions with relative errors smaller than 10% are accepted. This test can be used to detect major changes in the reconstructor that result in dimension mismatch or accuracy deviation from the ground true images.


camFoltz:
I like the idea of testing against simulated data, it allows for a bit less overhead in testing.  I tried everything out and looked at the intermediate images, it all seems to check out.  I will approve these changes

camFoltz (approved):

tests/reconstructor/test_2D_QLIPP_recon.py
```diff
@@ -0,0 +1,131 @@
...
+
+    # phase_Tikhonov (sensitive to regularization tuning, might cause error if the regularization is not tuned properly)
+    assert(np.sum(np.abs(phase_gt-(phi_sample-np.mean(phi_sample)))**2)/np.sum(np.abs(phase_gt)**2) < 0.1)
+
+    # phase_TV
+    assert(np.sum(np.abs(phase_gt-(phi_sample_TV-np.mean(phi_sample_TV)))**2)/np.sum(np.abs(phase_gt)**2) < 0.1)
+
+    # retardance + orientation (no deconvolution)
+    assert(np.sum(np.abs(S1_gt-Recon_para[0,:,:,2]*np.sin(2*Recon_para[1,:,:,2]))**2)/np.sum(np.abs(S1_gt)**2) < 0.1)
+    assert(np.sum(np.abs(S2_gt-Recon_para[0,:,:,2]*np.cos(2*Recon_para[1,:,:,2]))**2)/np.sum(np.abs(S2_gt)**2) < 0.1)
```
bryantChhun, line 123: When would the `<0.1` condition be relevant?  Only any time you do Decon right?  Could the value be even less here?

Related too -- when you are doing decon, what is a reasonable difference between simulated and reconstructed?

bryantChhun (approved):
One question about the errors

tests/reconstructor/test_2D_QLIPP_recon.py
lihaoyeh: Thanks for bringing it up. This is a tricky question. Generally speaking, retardance measurement captures the DC component of the images but the phase measurement does not. Therefore, retardance measurement has overall lower error between the reconstructions and the ground true images. Potentially, we can use 0.5 % here for retardance. Here I get these numbers empirically by actually comparing the results from this current pipeline. Now 10% serves its function to detect major changes (bugs) when there is a major refactoring. We can iterate this next time we work on this test.

PR #20, opened 2021-05-21
fix incorrect assignment of channel, time, in one instance of the writer

camFoltz:
In one condition of the writer, the t and c channel were flipped.  fixed this bug


bryantChhun:
not sure why CI-pytests weren't launched.  I manually tested myself to confirm.

Will address the pytest issue in another PR

bryantChhun (approved):

Issue #25, opened 2021-05-26
acquisitions that are stopped early produce incorrect zarr outputs


bryantChhun:
## problem
A micro-manager acquisition that stops early (either from a crash or manually) does not update the ome.xml metadata, which is determined based on defined experiment parameters at the start of acquisition.  Consequently, `tifffile`, `zarr`, and our `io.MicromanagerReader` expect a larger dataset than exists.  The result is a written file that has a larger dimensionality than the source file.

## solution
Unclear right now.  It's possible the ome.tiff structure has something in place to signify "empty data" or similar...  Will need some research

bryantChhun:
Fixed in PR #28

PR #28, opened 2021-06-09
Interrupted acquisitions

bryantChhun:
This PR fixes issue #25 

The way I solve the problem for singlepage tiffs is different than for multipage tiffs:
**multipage tiffs**
I use ome metadata, as per Ivan's suggestion, which correctly reflects the final dimensionality

**singlepage tiffs**
The code already parses the coordinate-to-filename mapping from mm metadata.  That mapping reflects actual data --> files that definitely exist in the folder.  The code in this PR simply combs through all coordinates and extracts the number of each dimension from there.

I've also included simulated data (stopped early), and appropriate tests for that


camFoltz:
Do we want to implement acquisition that fail mid-position?  Such that they only acquire a few channels of that position?  And if so, we load the last completed timepoint/position?  We wouldprobably have to load the last completed timepoint in this case since it is the outermost dimension.  Having timepoints with different dimensions might be a headache for reconstruction?

I am open to merging this for now and revisiting later if you think that is best

bryantChhun:
> Do we want to implement acquisitions that fail mid-position, i.e. they only acquire a few channels of that position? And if so, we load the last completed timepoint/position? We would probably have to load the last completed timepoint in this case since it is the outermost dimension. Having timepoints with different dimensions might not be possible or would be a headache for reconstruction
> 
> I am open to merging this for now and revisiting later if you think that is best

Interesting point.  It's not clear to me.  I think cutting out that last frame makes sense from a processing standpoint, but not from a data-purity standpoint (you can easily cut out that last frame manually afterwards, but do we want to leave this to the user?)

If we decide to automatically cut out incomplete "frames", it should be easy enough to add a single function to check "completeness" every time point.

My feeling right now is to leave this for another PR

camFoltz (approved):
Discussed in person.  Tests and Changes look good.  Will leave mid-position failure acquisition for another PR

PR #29, opened 2021-06-10
add z_step_size to attributes from metadata

camFoltz:
add z_step_size to reader class attributes from metadata.


bryantChhun (approved):
looks good

PR #32, opened 2021-06-21
add padding

camFoltz:
add padding to wavelet denoising to account for the function only working with even dimensions.  Implements a fix for any nD array passed to the input


lihaoyeh (approved):
Thanks, Cam. The code looks great and I tested with uPTI datasets. It produces images with output dimensions consistent with the input dimensions when image dimension has odd numbers in it. There is no major artifacts coming out from it.

PR #35, opened 2021-06-30
Automatically choose regularization parameters with L-curve

tiffchien:


waveorder/util.py
```diff
@@ -871,52 +872,180 @@ def Dual_variable_ADMM_TV_deconv_2D(AHA, b_vec, rho, lambda_u, lambda_p, itr, ve
...
+
+        return 4*signed_area / (d0*d1*d2)
+    
+    # evaluate the L curve at a specific lambda
+        # returns a new Point_L_curve() tuple.
+        # this function is the only place where new points are instantiated
+    def eval_L_curve(reg_x):
+        reg_coeff = 10**reg_x
+        H_eff_abs_square = xp.abs(H_eff)**2
+        H_eff_conj = xp.conj(H_eff)
```
lihaoyeh, line 952: Line 951 and 952 can be shared for all L-curve evaluations, so maybe we can move them out of the `eval_L_curve` functions?

waveorder/util.py
```diff
@@ -871,52 +872,180 @@ def Dual_variable_ADMM_TV_deconv_2D(AHA, b_vec, rho, lambda_u, lambda_p, itr, ve
...
+
+        # FT{f} (f=scattering potential (whose real part is (scaled) phase))
+        f_real_f = S0_stack_f * H_eff_conj / (H_eff_abs_square + reg_coeff)
+        S0_est_stack_f = H_eff * f_real_f # Ax (put estimate through forward model)
+
+        data_norm_eval = xp.log(xp.linalg.norm(S0_est_stack_f - S0_stack_f)**2 /N/M/L)
+        reg_norm_eval = xp.log(xp.linalg.norm(f_real_f)**2 /N/M/L)
+
```
lihaoyeh, line 960: Line 956--960 are only used for evaluating points on L-curve but are not used for reconstruction without auto-tuning. Maybe consider creating a function specifically for reconstruction and the other function for L-curve evaluation that will call the reconstruction function when calculating `f_real_f`?

waveorder/util.py
```diff
@@ -871,52 +872,180 @@ def Dual_variable_ADMM_TV_deconv_2D(AHA, b_vec, rho, lambda_u, lambda_p, itr, ve
...
+        pt_right = eval_L_curve(opt_list[-1].reg_x + epsilon_auto)
+        f_real.append(ifft_f_real(pt_left.f_real_f))
+        f_real.append(ifft_f_real(opt_list[-1].f_real_f))
+        f_real.append(ifft_f_real(pt_right.f_real_f))
         
-    return f_real
+        return np.array(f_real)
+    
+    else:
+        chosen_pt = eval_L_curve(xp.log10(reg_re))
```
lihaoyeh, line 1044: When no auto-tuning, we do not need to evaluate the coordinate on L-curve. This can save a bit of computation for non-autotuning situation.

waveorder/util.py
```diff
@@ -871,52 +872,180 @@ def Dual_variable_ADMM_TV_deconv_2D(AHA, b_vec, rho, lambda_u, lambda_p, itr, ve
...
-        S0_stack_f = fftn(S0_stack, axes=(-3,-2,-1))
-        
-        f_real     = np.real(ifftn(S0_stack_f * np.conj(H_eff) / (np.abs(H_eff)**2 + reg_re),axes=(-3,-2,-1)))
+        if verbose:
+            print('Final regularization parameter chosen: %.2e' % opt_list[-1].reg)
+            
+        # Return 3 solutions: the parameter chosen by the L-curve +/- epsilon
+        f_real = []
+        pt_left = eval_L_curve(opt_list[-1].reg_x - epsilon_auto)
+        pt_right = eval_L_curve(opt_list[-1].reg_x + epsilon_auto)
```
lihaoyeh, line 1036: We only need reconstructions from `pt_left` and `pt_right`, so we can save the computation for L-curve evaluation.

waveorder/util.py
```diff
@@ -871,52 +872,180 @@ def Dual_variable_ADMM_TV_deconv_2D(AHA, b_vec, rho, lambda_u, lambda_p, itr, ve
...
+        reg_x[3] = reg_x_cent + search_range_auto
+        reg_x[1] = calc_golden_x(reg_x[0], reg_x[3])
+        reg_x[2] = reg_x[0] + (reg_x[3]-reg_x[1])
+
+        # holds the 4 current points
+        curr_pts = []
+        for i in range(4):
+            curr_pts.append(eval_L_curve(reg_x[i]))
+        
+        opt_list = [] # save optimal point from each iteration
```
lihaoyeh, line 990: We might consider saving each optimal points when we want to debug. For real application, if we do not use the history optimal points, we could consider not saving them to save some memory footprint.

lihaoyeh (commented):
I have tested this PR and it works for 3D phase and fluorescence deconvolution. The code is succinct and clear. The computation speed can be improved a bit after some refinement of the code. Let's make sure we optimize for the speed a bit more before merging the PR. Thanks, Tiffany!

camFoltz:
May be making a change in the auto-regularization to be compatible with recOrder napari plugin.  Will update this PR

camFoltz:
I have added one small feature to output lambda after the regularization, I will use this value for the reconstruction autoregularizer UI.

Just to confirm `last_opt.reg_x` is the regularization value correct? I.e. I could pass this value directly back into this function as `reg_re`?

waveorder/util.py
```diff
@@ -871,52 +872,202 @@ def Dual_variable_ADMM_TV_deconv_2D(AHA, b_vec, rho, lambda_u, lambda_p, itr, ve
...
+            
+        # Return 3 solutions: the parameter chosen by the L-curve +/- epsilon
+            # try to not keep intermediate values in memory
+        f_real = []
+        f_real.append(ifft_f_real(compute_f_real_f(last_opt.reg_x - epsilon_auto)))
+        f_real.append(ifft_f_real(compute_f_real_f(last_opt.reg_x)))
+        f_real.append(ifft_f_real(compute_f_real_f(last_opt.reg_x + epsilon_auto)))
+
+        if output_lambda:
+            return np.array(f_real), last_opt.reg_x
```
lihaoyeh, line 1061: Hey @camFoltz ,

last_opt.reg_x is log_10 of the regularization parameter. If you wish to use this number for reg_re, you would need to turn it this way: 
`reg_re = 10**(last_opt.reg_x)`
Otherwise, this parameter pullout looks good to me.

camFoltz: Sounds good, I just updated this to output the non-log scale.  Thanks!

camFoltz (approved):

lihaoyeh (approved):

PR #37, opened 2021-09-14
OMEZarr HCS Writer

camFoltz:
This should be our stable writer build now for all raw data / processed data.  recOrder pipeline will have to be updated to work with this new writer.  The current layout is to have one of two workers under the main writer module to handle HCS tiled-like data and the Default data (which is HCS format with one row and N columns for N positions).  

The HCSZarr writer needs a specific set of metadata passed to it in order to create the HCS structure.  A user can do this themselves following OME specification, or we can parse / create the metadata automatically (I will implement this in the converter / recOrder pipeline).  The motivation for this use case was for processing data through a pipeline, where the raw_data hierarchy/positional metadata can be translated over and used for saving processed data-- ensuring that the two structures match in the end (although the arrays will certainly be different).

There may be other API calls to add that make it a bit more usable, but please check out the usage in `/waveorder/examples/Reader_Writer_Examples/Writer_Example.ipynb`

cc @lihaoyeh 
cc @ieivanov 

this notebook can help clarify a script for converting non-standard data formats


tests/writer/test_writer_v2.py
```diff
@@ -0,0 +1,64 @@
+import pytest
+import zarr
+from waveorder.io.writer import WaveorderWriter
+
+def test_default_writer():
+
+
+    data_path = '/Users/cameron.foltz/Desktop/Test_Data/Writer_Testing_OME/B1_small_noRawData.zarr'
+    out_path = '/Users/cameron.foltz/Desktop/Test_Data/Writer_Testing_OME/Writer_Output/'
```
bryantChhun, line 9: will need to fix these for pytests.  Can simply use the temp folder construction/destruction model (for writing).  Maybe consider fake data generation in the conftests.py that is all in-memory.

bryantChhun (approved):
Overall I think it looks good.

In a future PR, it'll be worth thinking about the API -- specifically the position handling

- I'm not sure this concept of a "current position" is so intuitive.  Instead, you could ask the user to specify the position at _write time_, such as `writer.write(data[0], position=0, t=..., c=...)`

- array initialization:  Could this happen before `writer.open_position(#)`?  I understand that initialization is an expensive process for large datasets -- so maybe this should be done automatically (whenever a specific position is addressed) or manually (for each position ahead of time).

This overhead of initialization is a bit of a headache -- and I think this explicit position-selection procedure is a way of getting around that, but it makes the interaction a bit clunky.

tests/writer/test_writer_v2.py
camFoltz: > In a future PR, it'll be worth thinking about the API -- specifically the position handling

I fully agree here, I followed what was in place, but we can potentially simplify how it moves through positions and it might be more intuitive for the user.

> * I'm not sure this concept of a "current position" is so intuitive.  Instead, you could ask the user to specify the position at _write time_, such as `writer.write(data[0], position=0, t=..., c=...)`

Yes we could do that, the concept of creating / opening positions allows the user to define the structure.  We can probably update the structure as we write, i.e. call `open_position` and `create_position` upon write.  Would simplify the work the user has to do.  Will think about this more so the user can call more than a single position upon write.

> * array initialization:  Could this happen before `writer.open_position(#)`?  I understand that initialization is an expensive process for large datasets -- so maybe this should be done automatically (whenever a specific position is addressed) or manually (for each position ahead of time).

Technically it cannot as of now, the position subgroup / row / column hierarchy needs to be created before the user initializes the array.  We can potentially just couple this create into initialize array.  In this case, the above mentioned issue is then covered with the array initialization

> This overhead of initialization is a bit of a headache -- and I think this explicit position-selection procedure is a way of getting around that, but it makes the interaction a bit clunky.

Yeah I agree definitely room for improvement / simplification here.  Will address these in the next PR

camFoltz: I will whip up some tests

camFoltz:
Simplified the API per your suggestions and wrote some tests @bryantChhun can you take a quick look

bryantChhun (approved):

PR #38, opened 2021-09-15
Zarr reader

camFoltz:
Basic reader for our zarr HCS format.  Compliments the writer.  No bells and whistles just yet but the same API calls work for this reader as they did for the previous ones.  All parsing is based off of ome-zarr metadata


camFoltz:
Still need to write tests.  We don't have a good example zarr dataset yet

bryantChhun (approved):

PR #40, opened 2021-09-28
fix bug in MM meta

camFoltz:
small bug fix

### DIFF
--- a/waveorder/io/zarrfile.py
+++ b/waveorder/io/zarrfile.py
@@ -44,7 +44,10 @@ class ZarrReader(ReaderInterface):
 
         # initialize metadata
         self.mm_meta = None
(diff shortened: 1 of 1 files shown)


waveorder/io/zarrfile.py
```diff
@@ -44,7 +44,10 @@ def __init__(self,
 
         # initialize metadata
         self.mm_meta = None
-        self._set_mm_meta()
+        try:
+            self._set_mm_meta()
+        except:
```
bryantChhun, line 49: you typically want to catch specific exceptions:
```
except AttributeError as ae:
    logs.warning(f"{ae}")
    self.mm_meta = None
```

bryantChhun (approved):

PR #43, opened 2021-09-30
UPTI File Reader

camFoltz:
This creates a reader subclass that allows for reading of UPTI files.

Creates a lazy zarr store in memory with the UPTI channels (patterns, states)

Works with both live upti and original upti which has states present or not present.

I have it hardcoded right now that the channels will always go pattern then state. i.e. an acquisition with 9 patterns and 4 states would have the following channel order and channel names:
```
channel_names = ['pattern_000_state_000',
                 'pattern_000_state_001',
                 'pattern_000_state_002',
                 'pattern_000_state_003',
                 'pattern_001_state_000',
                 ...]
```

the placement of the channel in the array corresponds to the order of the list.

This reader will be used for the zarr converter and UPTI can now be converted to a zarr store on disk

### DIFF
--- a/waveorder/io/upti.py
+++ b/waveorder/io/upti.py
@@ -21,7 +21,7 @@ class UPTIReader(ReaderInterface):
 
         self.data_folder = folder
         self.files = glob.glob(os.path.join(folder, '*.tif'))
-        info_img = tiff.imread(self.files[0]).dtype
+        info_img = tiff.imread(self.files[0])
         self.dtype = info_img.dtype
         (self.height, self.width) = info_img.shape
         self.positions = 1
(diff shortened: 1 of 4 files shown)


waveorder/io/upti.py
```diff
@@ -0,0 +1,191 @@
...
+
+    def __init__(self, folder: str, extract_data: bool = False):
+
+        # check if folder exists
+        if not os.path.isdir(folder):
+            raise ValueError("folder does not exist")
+
+        # Initializate data parameters
+        self.data_folder = folder
+        self.files = glob.glob(os.path.join(folder, '*.tif'))
```
lihaoyeh, line 24: Some of my super old files are saved under ".tiff" rather than ".tif". An example is under `/gpfs/CompMicro/rawdata/hummingbird/LiHao/20200223_LCD_63x_147_oil_test/Mouse_brain_702_ROI2/data_2/`. It will throw error because of this mismatch. Potentially an `if` statement may do it okay.

camFoltz:
> The SONY polarization camera takes all four polarization channels at the same time. We can potentially have the channel assignment happened within the reader somehow.

We can look for the Sony Camera in the device property metadata and automatically split the pixels into the 4 polarization states.  That is a nice feature to add!

> Since each uPTI dataset takes up much memory space, it would be great to have an option to only load a partial FOV of data for processing.

I also agree with this.  I implemented a similar feature for the multi-page tiff.  I will make this update!

PR #48, opened 2021-12-14
Fix bug in reading data with channels that don't have z stack

ieivanov:
MM allows you to not collect z stacks for certain channels. This commit allows you to read such data. Data from channels with only a single slice is placed at position num_slices//2 in the array

### DIFF
--- a/waveorder/io/multipagetiff.py
+++ b/waveorder/io/multipagetiff.py
@@ -247,10 +247,9 @@ class MicromanagerOmeTiffReader(ReaderInterface):
                                                dtype=self.dtype)
         # add all the images with this specific dimension.  Will be blank images if dataset
         # is incomplete
-        for t in range(timepoints):
-            for c in range(channels):
-                for z in range(slices):
-                    self.position_arrays[pos][t, c, z, :, :] = self.get_image(pos, t, c, z)
+        for p, t, c, z in self.coord_map.keys():
+            if p == pos:
+                self.position_arrays[pos][t, c, z, :, :] = self.get_image(pos, t, c, z)
 
     def _set_dtype(self):
         """


waveorder/io/multipagetiff.py
```diff
@@ -247,10 +247,9 @@ def _create_position_array(self, pos):
                                                dtype=self.dtype)
         # add all the images with this specific dimension.  Will be blank images if dataset
         # is incomplete
-        for t in range(timepoints):
-            for c in range(channels):
-                for z in range(slices):
-                    self.position_arrays[pos][t, c, z, :, :] = self.get_image(pos, t, c, z)
+        for p, t, c, z in self.coord_map.keys():
```
camFoltz, line 250: @ieivanov In the case of a z-stack not acquired for a specific channel, is the coordinate from the IndexMap metadata for this channel the middle z-slice?

i.e. If Pos0, Time0, Channel 1 does not have a z-stack, but the rest of the channels have 81 slices does the coordinate look like (P, T, C, Z) = (0, 0, 1, 40) in the metadata?

camFoltz (approved):
This change looks good, I will have to update the zarr tests as it looks like they can no longer find the example datasets

waveorder/io/multipagetiff.py
ieivanov: That's correct. The way the zarr array is structured right now, you will get a zeros image if you ask for anything but the middle slice, which is the expected behavior.

camFoltz (approved):
Good to merge

waveorder/io/multipagetiff.py
camFoltz: Cool, thanks for catching this!

PR #49, opened 2021-12-14
PyPi Release Prep

camFoltz:
Setting up the setup.py file for release


lihaoyeh (approved):
I have tested this PR on IBM machine. The current `setup.py` will ignore package requirements on PowerPC. It will not try to install anything so it does not error out anymore. It is expected that people who would like to use `waveorder` on PowerPC will need to build their environments manually. At the same time, the setup.py will install necessary packages on regular x64 machines for proper deployment.

PR #55, opened 2022-01-03
Instrument tensor reconstruction

ieivanov:
Instrument tensor recon can be tested with this notebook: https://github.com/czbiohub/compmicro-sandbox/blob/master/Ivan/Python%20Scripts/MI-PolScope%20reconstruct%20data%20Fig1.ipynb


waveorder/waveorder_reconstructor.py
```diff
@@ -977,23 +988,40 @@ def Stokes_recon(self, I_meas):
...
+            raise ValueError('Unsupported image data size. Image data must be of size (N_channel, N, M) or '
+                             '(N_channel, N, M, N_defocus)')
+
+        if n_dims == 3:
+            I_meas = I_meas[..., np.newaxis]
+        img_data = np.moveaxis(I_meas, 0, -2)
+
+        if self.use_gpu:
+            if self._A_matrix_inv_gpu_array is None:
+                self._A_matrix_inv_gpu_array = cp.array(self.A_matrix_inv)
```
camFoltz, line 1014: Question -- do we need to keep the `_A_matrix_inv_gpu_array` as a class attribute?  It doesn't look like it is ever called again outside of this function.  Perhaps we can leave it as a temporary variable within this function?

camFoltz (commented):
Overall everything looks good and seems to test fine.  This will simplify the reconstruction pipeline in recOrder a bit as well.  Thanks for making these changes!

waveorder/waveorder_reconstructor.py
ieivanov: I reuse the `_A_matrix_inv_gpu_array` in repeated calls to `Stokes_recon`, but it's true that it is not used by other methods. Do you have another way to handle that?

camFoltz: Ok yeah this makes sense then, I think this is a fine way of handling it.

camFoltz (dismissed):
This looks good to me.  I have tested the `stokes_recon` in both the 2D and 3D case and it works well.

waveorder/waveorder_reconstructor.py
```diff
@@ -977,23 +988,40 @@ def Stokes_recon(self, I_meas):
...
-        
-        A_pinv = np.linalg.pinv(self.A_matrix)
-        S_image_recon = np.reshape(np.dot(A_pinv, I_meas.reshape((self.N_channel, -1))), (self.N_Stokes,)+img_shape[1:])
 
-            
-        return S_image_recon
+        n_dims = I_meas.ndim
+        if n_dims not in (3, 4):
+            raise ValueError('Unsupported image data size. Image data must be of size (N_channel, N, M) or '
+                             '(N_channel, N, M, N_defocus)')
```
lihaoyeh, line 1006: In uPTI reconstruction pipeline, my intensity data is usually structured as (N_channel, N_pattern, N, M) for 2D imaging and (N_channel, N_pattern, N, M, N_defocus) for `Stokes_recon` because the code is used to support N-D processing. 

Could we keep the original implementation for N-D processing and conditionally have the instrument-tensor processing activated when `self.A_matrix_inv` has `ndim = 4`?

ieivanov: Thanks @lihaoyeh, I'll rework this implementation. I also saw that in recOrder @camFoltz passes XYZT stacks to this function, and that throws an error too. 

@lihaoyeh can you please include a test for uPTI data?

ieivanov:
I think Li-Hao's comment touches on a longer discussion on how data should be structured. During reconstruction, having consistent dimensions (e.g. `XYZCT`) will eliminate the need for handling special cases (such as 2D images, Z stacks, or time series) and will make the code cleaner.

I would propose that the reconstruction unit of waveorder should be a Z stack (i.e. `ZYX` or `YXZ` dimension order). If we want to reconstruct a 3D plane we'd pass an array with size `(1, sizeY, sizeX)`. If we want to reconstruct a time series, then we'd iterate over the time dimension, calling the reconstruction functions once per time point.

@lihaoyeh in the case of uPTI data you still may be able to structure the data in `XYZCT` format with the channel dimensions running over Stokes channels and illumination patters in defined order (e.g. I0-P0, I0-P1, ... I1-P0, ... In-Pm). In that case you'd have an instrument matrix of matching dimensionality - (n*m) x 4, where each of the m rows is the came. Or you can call reconstruct_Stokes multiple times for each illumination patters. 

This would require some restructuring of the current code, but I think it would make it cleaner. Let me know what you guys think.

@lihaoyeh @camFoltz @mattersoflight

waveorder/waveorder_reconstructor.py
lihaoyeh: That's a good idea. I can work on an integrated test for uPTI simulation similar to what's in `waveorder/tests/reconstructor/test_2D_QLIPP_recon.py`. The current bottleneck is to figure out how to run this test with minimum computing resources because the full forward/inverse simulation for uPTI can be computationally intense. Before I finish that, you can temporarily run the notebooks in `waveorder/examples/uPTI_simulation/` for testing. Those notebooks should at least provide the check on the array dimensions. We will have to rely on the review, for now, to identify whether there could be potential accuracy issues.

camFoltz: > Thanks @lihaoyeh, I'll rework this implementation. I also saw that in recOrder @camFoltz passes XYZT stacks to this function, and that throws an error too.

@ieivanov do you mean XYZC here?  I do not ever pass an array with a time dimension XYZT or XYZCT into any of these functions, my only use case is at the level of image volumes.

waveorder/waveorder_reconstructor.py
```diff
@@ -977,23 +988,40 @@ def Stokes_recon(self, I_meas):
...
+            img_gpu_array = cp.array(img_data)
+            S_image_recon = cp.asnumpy(cp.matmul(self._A_matrix_inv_gpu_array, img_gpu_array))
+        else:
+            S_image_recon = np.matmul(self.A_matrix_inv, img_data)
+
+        S_image_recon = np.moveaxis(S_image_recon, -2, 0)
+        if n_dims == 3:
+            return S_image_recon[..., 0]
+        else:
+            return S_image_recon
```
mattersoflight, line 1024: A potential solution that can handle both instrument matrix and instrument tensor cases is to vectorize only XZ dimensions (i=Z*X*Y) in intensity, stokes, and instrument tensor arrays. In this case, intensities will be of (i,4) shape for both 2D images and 3D stacks. Correspondingly, Stokes array will be (i,3) shape. The inverse instrument tensor can have the shape (1,4,3) OR (i,4,3). 

Based on the broadcasting behavior of numpy.matmul (noted below), it seems like np.matmul or cp.matmul will multiply arrays of shape (i,4) and (i,4,3) to return an array of shape (i,3). 

(https://numpy.org/doc/stable/reference/generated/numpy.matmul.html, "If either argument is N-D, N > 2, it is treated as a stack of matrices residing in the last two indexes and broadcast accordingly.")

mattersoflight:
We should resume the discussion of the optimal dimension order. But, it is best to do this refactor in a dedicated PR. For now, let's retain the backward compatibility while incorporating instrument tensor.

waveorder/waveorder_reconstructor.py
mattersoflight: This could work
```
import numpy as np
nVox=100
inst_mat=np.array([[0.25,0.25,0.25,0.25],[1,0,-1,0],[0,1,0,-1]])
inst_mat=inst_mat.reshape(1,3,4)
print("inst_mat:")
print(inst_mat)

inst_tensor=np.repeat(inst_mat,nVox,axis=0)
intensity=np.random.rand(nVox,4,1)
print("instrument tensor:"+str(inst_tensor.shape))
print("intensity shape:"+ str(intensity.shape))
stokes = np.matmul(inst_tensor, intensity)
print("stokes shape:" + str(stokes.shape))
```

produces

```
inst_mat:
[[[ 0.25  0.25  0.25  0.25]
  [ 1.    0.   -1.    0.  ]
  [ 0.    1.    0.   -1.  ]]]
instrument tensor:(100, 3, 4)
intensity shape:(100, 4, 1)
stokes shape:(100, 3, 1)
```

ieivanov: Thanks Shalin, the issue with the current implementation is that the function needs to accept data of formats `N_channel, N, M`, `N_channel, N_pattern, N, M`, `N_channel, N, M, N_defocus`, `N_channel, N_pattern, N, M, N_defocus`, and its ambiguous where the image dimensions are. The previous implementation used to reshape I_mean into `(N_channel, -1)` and then multiply by the instrument matrix, this strategy does not work in the case of the instrument tensor

ieivanov:
@lihaoyeh @camFoltz @mattersoflight please take another look at this PR. I've reworked the `Stokes_recon` function, it now works for instrument tensor reconstruction, uPTI reconstruction, and instrument matrix reconstruction (built into the tests).

I plan to add the fluorescence anisotropy recon function to the `fluorescence_microscopy` class before we merge this PR, most likely tomorrow.

lihaoyeh:
@ieivanov Thanks for reworking the code to keep the backward compatibility. I have tested `Stokes_recon` in a few of my reconstruction notebooks. They work as expected. Once you have the fluorescence anisotropy recon written, I am happy to test it with some of your old data. Thank you again for doing this.

ieivanov:
I've added the `Fluor_anisotropy_recon` to the `fluorescence_microscopy` class, @mattersoflight please take a look when you have a chance.

However, my pytests fail, it seems like the tester cannot find the `2T_3P_81Z_231Y_498X_Kazansky.zarr`, @camFoltz can you please take a look?

camFoltz:
> I've added the `Fluor_anisotropy_recon` to the `fluorescence_microscopy` class, @mattersoflight please take a look when you have a chance.
> 
> However, my pytests fail, it seems like the tester cannot find the `2T_3P_81Z_231Y_498X_Kazansky.zarr`, @camFoltz can you please take a look?

For some reason GitHub has trouble sometimes with the GoogleDriveDownloader where it doesn’t fully finish grabbing the data.  You just have to re-run the tests and it usually works the second time!

I’ll test out the Stokes_recon function as well.

camFoltz (dismissed):
These changes integrate well with the recOrder wrapper functions and normal QLIPP compute workflow.  I didn't run into any issues here.

camFoltz:
I did add `deconv_mode` to the `fluorescence_microscopy` class attributes so recOrder can keep track of whether the reconstructor has changed from 2D to 3D (in order to see if it needs to recompute the reconstructor)

lihaoyeh:
Hey @ieivanov,
I will approve this PR once the master branch is merged into this branch without conflicts.

lihaoyeh (approved):
I have tested this PR again. It is compatible with the pipeline used for uPTI reconstruction.

ieivanov:
This branch successfully passes all tests, I think it's ready to merge with the main. @lihaoyeh can you approve?

lihaoyeh (approved):

PR #73, opened 2022-04-05
multiprocess script refactoring, set default to run full job

JohannaRahm:
changed the default behavior of the phase & deconvolution multiprocessing [script](https://github.com/mehta-lab/waveorder/blob/waveorder_scripts/examples/multiProcessing/Phase3D_fluor_deconv_waveorder_all_channels.py) to run the full job and refactored it (more comments, all variables that must be set by user on top).


talonchandler (approved):
Looks good to me! Thanks for updating Johanna.

PR #77, opened 2022-06-28
Fix ome-tif to zarr converter for acquisitions with micromanager beta

talonchandler:
This PR is parallel to [this recOrder PR](https://github.com/mehta-lab/recOrder/pull/123), which [addresses this issue](https://github.com/mehta-lab/recOrder/issues/121). I will merge both PRs at the same time.


JohannaRahm (approved):
reminder to take the except KeyError version when merging. Both PRs look good - thanks again for fixing the issue in such a timely manner.

PR #79, opened 2022-07-13
Pycromanager data reader for WaveorderReader class

ieivanov:
This PR extends the WaveorderReader with a subclass for reading pycromanager data with regular dimensions. Reading pycromanager data with irregular dimensions may not be supported, to be tested and extended at a later point.

In an accompanying PR, the `recOrder` zarr converter is also extended to allow converting pycromanager data into a zarr store that can be easily loaded in `napari`. Currently viewing pycromanager data is difficult, even if it has regular dimensions.

Note: The `Dataset.as_array()` function in pycromanager 0.13.2 loads the data into memory, which is not the intended behavior of the `WaveorderReader.get_zarr()` wrapper for this function. This bug is fixed in the current (0.18.3) version of pycromanager - with this version `WaveorderReader.get_zarr()` returns a lazy-loaded *Dask* array. I haven't been able to cast the dask array into zarr without bringing the data into memory, I think returning a dask array is OK for now. I think this is one good reason to upgrade to a move recent version of pycromanager (and MM).


ieivanov:
@talonchandler this PR is ready for review and merge.

mehta-lab/recOrder#130 depends on a pypi release of this version of `waveorder`. Here is our action plan:
* approve this PR for merge
* resolve mehta-lab/recOrder#88 and upload new test dataset to Zenodo
* update test dataset url in `waveorder` and `recorder`; waveorder test dataset will be unchanged apart from the new url
* merge this PR
* release new version of `waveorder` on pypi
* merge mehta-lab/recOrder#130 

Let me know what you think

talonchandler:
Plan sounds good! I'll review this this afternoon.

waveorder/io/reader.py
```diff
@@ -53,52 +54,20 @@ def __init__(self,
...
+            elif WaveorderReader._check_pycromanager(src):
+                data_type = 'pycromanager'
+            elif WaveorderReader._check_multipage_tiff(src):
+                data_type = 'ometiff'
+            elif WaveorderReader._check_single_page_tiff(src):
+                data_type = 'singlepagetiff'
+            else:
+                raise FileNotFoundError(f'No compatible data found under {src}, please specify the top '
+                                        'level micromanager directory.')
+        self.data_type = data_type
```
talonchandler, line 70: Very nice!

waveorder/io/reader_base.py
```diff
@@ -1,8 +1,25 @@
...
 
+
 class ReaderBase:
 
+    def __init__(self):
+        self.frames = None
+        self.channels = None
+        self.slices = None
+        self.height = None
+        self.width = None
```
talonchandler, line 12: These lines, together with moving the `.shape` method here, makes sense to me.

waveorder/io/reader_base.py
```diff
@@ -1,8 +1,25 @@
...
+        self.dtype = None
+        self.mm_meta = None
+        self.stage_positions = None
+        self.z_step_size = None
+        self.channel_names = None
```
talonchandler, line 17: I'm less clear about the motivation for these lines since they're not used in this parent class? Is is just so that you can be sure that all of the readers have something in these fields?

talonchandler (dismissed):
Only minor questions about the class structure...thanks for doing this!

talonchandler:
Thanks for flagging the problem with lazy-loading in pycromanager 0.13.2. I think a pycromanager upgrade to 0.18.3 makes sense for recOrder 1.0.0, so I'll write a recOrder issue (since we're planning to move io to recOrder for 1.0.0). I'll tag this PR so that we don't forget.

ieivanov:
Thanks Talon!

waveorder/io/reader_base.py
ieivanov: I don't fully remember, but I think this means that we expect all readers which inherit this base class to have these attributes. We check for these attributes in our tests

talonchandler (approved):

talonchandler:
Ready to merge from my perspective. 

We want this version pushed to pypi, right? I can do this today, and I'll name it 1.0.0b1.

talonchandler:
I can push to pypi version 1.0.0b1 when I get back (Aug 22), but if this is holding you up feel free to go ahead and push yourself. 

**Instructions:**
1. Manually update the version number in `setup.py` to `1.0.0-beta1` (we're using `setuptools_scm` and git tags in `recOrder`, but I haven't set this up in`waveorder`).
2. Login to PyPI with compmicro's lastpass.
3. Use PyPI to generate a token and put it in your `.pypirc` file like this
```
[pypi]
username = __token__
password = pypi-*******
```
4. Navigate to the `waveorder` folder and checkout the version you'd like to push. Use the following commands: 
```
python setup.py sdist bdist_wheel # generates a tarball and wheel. 
# Optional: use `pip install dist/*whl` to test the wheel that will be uploaded.
pip install --upgrade twine # install twine
twine check dist/* # check that the tarball and wheel are ready for upload
twine upload dist/* # uploads to PyPI, requires key in .pypirc
```

PR #84, opened 2022-08-29
BSD 3-clause

talonchandler:
Updating `setup.cfg` metadata. This field isn't widely used, but should be synchronized to avoid any confusion.


mattersoflight (approved):
thanks

PR #86, opened 2022-09-15
Pin pywavelets version

talonchandler:
`pywavelets` latest release doesn't have working wheels, which is interrupting my testing. I'm pinning `pywavelets` for now so that we can continue. 

When this `pywavelets` issue is resolved we can consider unpinning: 
https://github.com/PyWavelets/pywt/issues/656

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -35,10 +35,10 @@ python_requires = >3.7
 setup_requires = setuptools_scm
 # add your package requirements here
 install_requires =
+    pywavelets==1.1.1
     numpy>=1.17.4
     matplotlib>=3.1.1
     scipy>=1.3.0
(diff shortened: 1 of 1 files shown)


talonchandler:
I'll hold off on merging this for now, since it'll likely be fixed in the next few days. For now we can run

`pip install pywavelets==1.1.1` before running 

`pip install waveorder` or `pip install recOrder-napari`.

ziw-liu:
It seems like they [fixed the issue within hours](https://github.com/PyWavelets/pywt/issues/656#issuecomment-1249042476)!

talonchandler:
Thanks for checking! I just tested a fresh install and `pip install` now works smoothly installs `pywavelets==1.4.1`. Closing.

PR #96, opened 2022-12-02
Added example for multiple input deconvolution

Soorya19Pradeep:
Example script for deconvolution of multiple ometiff inputs to write single zarr output. The example performs phase reconstruction from BF and deconvolution of two fluorescence channels.

### DIFF
--- a/examples/multiProcessing/OmeTiff_3DDeconvolution_Zarr_waveorder.py
+++ b/examples/multiProcessing/OmeTiff_3DDeconvolution_Zarr_waveorder.py
@@ -1,0 +1,112 @@
+# script to take multiple ometiff input folders, perform deconvolutions for each channel, and write a single 
+# omezarr output. The current script performs phase reconstruct from 'BF' channel and deconvolution of two 
+# fluorescence channels 'DAPI' and 'Y5'. 
+
+# %% import libraries
+import os 
+import numpy as np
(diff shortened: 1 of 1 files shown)


talonchandler:
Thanks @Soorya19Pradeep. 

For additional context, this "multiset input" reconstruction takes positions from multiple MDAs (multiple folders each with multiple .ome.tifs) and outputs a single .zarr reconstruction. 

@ziw-liu do you think it's best to support "multiset" i/o on the `iohub` side? I'm leaning towards "yes" since Soorya mentioned that she and her collaborators commonly work in this mode of operation---splitting up a large number of positions into a smaller number of MDAs. If you agree, then you can use the i/o in this script as inspiration, and I think it's best not to merge this here. 

Alternatively, we can merge this for now and leave `iohub` handling until later. I don't this needs to block an initial release of `iohub`.

ziw-liu:
> support "multiset" i/o on the iohub side

This is definitely a valid feature request. The OpenCell team also generates multiple-MDA (OME-Tiff) datasets.

talonchandler:
Okay great thanks @ziw-liu! I'll close this for now. 

@Soorya19Pradeep thanks a bunch for sharing this script. If you need support with this type of io while `iohub` is spinning up...let us know.

PR #105, opened 2023-02-22
README copy edits

talonchandler:
@ziw-liu I'm preparing for the `waveorder==0.1.0rc1` release, and this PR includes two small edits to prepare for the release. 

Would you mind reviewing this PR? If you have 10 minutes, I'd also appreciate your copy edits on the README and anything else that you think might be required before release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ We provide an [example notebook](examples/fluorescence_deconvolution/) for widef
 
 ## Citation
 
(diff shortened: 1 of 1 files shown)


ziw-liu (approved):
LGTM.

On a side note, the equation in the figure is on a solid white background, while the rest of the figure is on a transparent background that may have contrast issues in a dark theme. But not within the scope of this PR.

PR #110, opened 2023-02-27
New Stokes and Mueller module

talonchandler:
This PR adds a completely rewritten version of all Stokes- and Mueller-related calculations in `waveorder`.

This rewrite was motived by the following questions:
- Q: how should we handle background correction of S3 to prepare it for deconvolution? A: keep track of S3 and use Mueller matrices to correct backgrounds. 
- how should we handle normalization (see the closed #103 for an earlier discussion)? A: no need to normalize in a separate step. 
- can we handle larger background retardances? A: yes, by estimating Mueller matrices. 

**Highlight improvements:**
- removal of normalization steps...instead of a two step reconstruction with a normalization followed by reconstruction, this implementation goes straight from Stokes parameters to (retardance, orientation, transmittance, depolarization). 
- a natively ND implementation...the Stokes and Mueller indices go in the first/second axes, and the remaining indices are arbitrary. A convenience `mmul` (Mueller multiply) function that uses `einsum` is the key simplifying design choice. 
- A Mueller-matrix based reconstruction scheme that can handle larger background retardances. 

**What does the new API look like?** Here's an example snippet from `/hpc/projects/compmicro/projects/infected_cell_imaging/Image_preprocessing/Exp_2023_02_07_A549MemNucl_PolPhase3D/Background_correction_trial/bg-corr-with-mask.py`
```
# Calculate A and A_inv
A = stokes.A_matrix(swing=0.1, scheme="5-State")
A_inv = np.linalg.pinv(A)

# Apply A_inv to background and sample data
S_bg = stokes.mmul(A_inv, cyx_bg_data)
S_sm = stokes.mmul(A_inv, czyx_data)

# Calculate background correction matrix from S_bg
M_inv = stokes.inv_AR_mueller_from_CPL_projection(*S_bg)

# Apply background correction to sample data
bg_corr_S_sm = stokes.mmul(M_inv, S_sm)

# Reconstruct parameters
ret, ori, tra, dop = stokes.inverse_s0123_CPL_after_ADR(*bg_corr_S_sm)
```

**Limitations compared to the current `waveorder_reconstructor` implementation:**
- No GPU implementation. @ziw-liu maybe you have ideas for flipping a gpu switch for this whole module? The `waveorder_reconstructor` class' parallel `np` and `cp` implementations seem clunky. 
- Not yet optimized...instead of using differences and ratios to apply background corrections, this implementation uses Mueller matrices and their inverses. This implementation is not slower than the phase reconstructions, and I've added comments in the places where further optimization can improve performance. 

I have not removed the existing implementation in the `waveorder_reconstructor` class. My current plan is to discuss the technical parts of this reimplementation and compare with the existing implementation here, then later I can complete the refactor by removing the Stokes parts of the `waveorder_reconstructor` class and updating the `recOrder` calls. 

Note: this PR is to merge into `alg-dev`, so we have a bit more flexibility in the changes. Temporarily breaking changes/suggestions are okay while we iterate. 

**Specific feedback requests:** 
- @ziw-liu your comments on a gpu switch, and on documentation+testing practice is very welcome. I wasn't sure if I should use type annotations & numpy-style docstrings? I stayed with only docstrings for now. 
- @mattersoflight I'm particularly interested in your thoughts on naming. For example, `inverse_s0123_CPL_after_ADR` doesn't roll off the tongue like the earlier `Polarization_recon`, but I think it's important to be very specific at this level. Later layers of abstraction can reintroduce more abstract names likes `Polarization_recon` if we think they're helpful.

### DIFF
--- a/tests/test_stokes.py
+++ b/tests/test_stokes.py
@@ -40,47 +40,45 @@ def test_stokes_recon():
 
                 # Test attenuating retarder (AR) functions
                 AR = (ret, ori, tra)
-                s012 = stokes.s012_CPL_after_AR(*AR)
-                AR1 = stokes.inverse_s012_CPL_after_AR(*s012)
+                s012 = stokes.stokes012_after_AR(*AR)
+                AR1 = stokes.estimate_AR_from_stokes012(*s012)
                 for i in range(3):
                     assert AR[i] - AR1[i] < 1e-8
 
                 # Test attenuating depolarizing retarder (ADR) functions
                 for dop in np.arange(1e-3, 1, 0.1):
                     ADR = (ret, ori, tra, dop)
-                    s0123 = stokes.s0123_CPL_after_ADR(*ADR)
-                    ADR1 = stokes.inverse_s0123_CPL_after_ADR(*s0123)
+                    s0123 = stokes.stokes_after_ADR(*ADR)
+                    ADR1 = stokes.estimate_ADR_from_stokes(*s0123)
 
                     for i in range(4):
                         assert ADR[i] - ADR1[i] < 1e-8
 
 
-def test_s0123_CPL_after_ADR_usage():
-    x = stokes.s0123_CPL_after_ADR(1, 1, 1, 1)
+def test_stokes_after_ADR_usage():
+    x = stokes.stokes_after_ADR(1, 1, 1, 1)
 
     ret = np.ones((2, 3, 4, 5))
     ori = np.ones((2, 3, 4, 5))
     tra = np.ones((2, 3, 4, 5))
     dop = np.ones((2, 3, 4, 5))
-    x2 = stokes.s0123_CPL_after_ADR(ret, ori, tra, dop)
+    x2 = stokes.stokes_after_ADR(ret, ori, tra, dop)
 
     ADR_params = np.ones(
         (4, 2, 3, 4, 5)
     )  # first axis contains the Stokes indices
-    stokes.s0123_CPL_after_ADR(*ADR_params)  # * expands along the first axis
+    stokes.stokes_after_ADR(*ADR_params)  # * expands along the first axis
 
 
-def test_AR_mueller_from_CPL_projection():
+def test_mueller_from_stokes():
     # Check thank inv(M) == M.T, (only true when
 
-    M = stokes.AR_mueller_from_CPL_projection(
+    M = stokes.mueller_from_stokes(
         1, 1 / np.sqrt(3), 1 / np.sqrt(3), 1 / np.sqrt(3)
     )
     assert np.max(np.linalg.inv(M) - M.T) < 1e-8
 
-    M2 = stokes.AR_mueller_from_CPL_projection(
-        1, 1 / np.sqrt(2), 1 / np.sqrt(2), 0
-    )
+    M2 = stokes.mueller_from_stokes(1, 1 / np.sqrt(2), 1 / np.sqrt(2), 0)
     assert np.max(np.linalg.inv(M2) - M2.T) < 1e-8
 
 
--- a/tests/test_stokes.py
+++ b/tests/test_stokes.py
@@ -4,17 +4,25 @@ import pytest
 import numpy.testing as npt
 
 
-def test_A_matrix():
-    A5 = stokes.A_matrix(0.1)
-    assert A5.shape == (5, 4)
+def test_S2I_matrix():
+    S2I5 = stokes.S2I_matrix(0.1)
+    assert S2I5.shape == (5, 4)
 
-    A4 = stokes.A_matrix(0.1, scheme="4-State")
-    assert A4.shape == (4, 4)
+    S2I4 = stokes.S2I_matrix(0.1, scheme="4-State")
+    assert S2I4.shape == (4, 4)
 
-    npt.assert_almost_equal(stokes.A_matrix(0), stokes.A_matrix(1))
+    npt.assert_almost_equal(stokes.S2I_matrix(0), stokes.S2I_matrix(1))
 
     with pytest.raises(ValueError):
-        Ax = stokes.A_matrix(0.1, scheme="3-State")
+        A2Ix = stokes.S2I_matrix(0.1, scheme="3-State")
+
+
+def test_I2S_matrix():
+    I2S5 = stokes.I2S_matrix(0.1)
+    assert I2S5.shape == (4, 5)
+
+    I = np.matmul(stokes.I2S_matrix(0.1), stokes.S2I_matrix(0.1))
+    npt.assert_almost_equal(I, np.eye(I.shape[0]))
 
 
 def test_s12_to_ori():
--- a/tests/test_stokes.py
+++ b/tests/test_stokes.py
@@ -85,7 +85,3 @@ def test_mmul():
     with pytest.raises(ValueError):
         M2 = np.ones((3, 4, 1))
         y2 = stokes.mmul(M2, x)
-
-    with pytest.raises(ValueError):
-        M3 = np.ones((3, 2, 1, 2))
-        y3 = stokes.mmul(M3, x)
--- a/tests/test_stokes.py
+++ b/tests/test_stokes.py
@@ -10,6 +10,8 @@ def test_A_matrix():
     A4 = stokes.A_matrix(0.1, scheme="4-State")
     assert A4.shape == (4, 4)
 
+    assert np.max(stokes.A_matrix(0) - stokes.A_matrix(1)) < 1e-8
+
     with pytest.raises(KeyError):
         Ax = stokes.A_matrix(0.1, scheme="3-State")
 
@@ -42,6 +44,21 @@ def test_stokes_recon():
 
                     for i in range(4):
                         assert ADR[i] - ADR1[i] < 1e-8
+
+
+def test_s0123_CPL_after_ADR_usage():
+    x = stokes.s0123_CPL_after_ADR(1, 1, 1, 1)
+
+    ret = np.ones((2, 3, 4, 5))
+    ori = np.ones((2, 3, 4, 5))
+    tra = np.ones((2, 3, 4, 5))
+    dop = np.ones((2, 3, 4, 5))
+    x2 = stokes.s0123_CPL_after_ADR(ret, ori, tra, dop)
+
+    ADR_params = np.ones(
+        (4, 2, 3, 4, 5)
+    )  # first axis contains the Stokes indices
+    stokes.s0123_CPL_after_ADR(*ADR_params)  # * expands along the first axis
 
 
 def test_AR_mueller_from_CPL_projection():
--- a/tests/test_stokes.py
+++ b/tests/test_stokes.py
@@ -12,7 +12,7 @@ def test_A_matrix():
 
     assert np.max(stokes.A_matrix(0) - stokes.A_matrix(1)) < 1e-8
 
-    with pytest.raises(KeyError):
+    with pytest.raises(ValueError):
         Ax = stokes.A_matrix(0.1, scheme="3-State")
 
 
--- a/tests/test_stokes.py
+++ b/tests/test_stokes.py
@@ -92,3 +92,17 @@ def test_mmul():
     with pytest.raises(ValueError):
         M2 = np.ones((3, 4, 1))
         y2 = stokes.mmul(M2, x)
+
+
+def test_copying():
+    a = np.array([1, 1])
+    b = np.array([1, 1])
+    c = np.array([1, 1])
+    d = np.array([1, 1])
+    s0, s1, s2, s3 = stokes.stokes_after_ADR(a, b, c, d)
+    s0[0] = 2  # modify the output
+    assert c[0] == 1  # check that the input hasn't changed
+
+    M = stokes.mueller_from_stokes(a, b, c, d)
+    M[0, 0, 0] = -1  # modify the output
+    assert a[0] == 1
--- a/tests/test_stokes.py
+++ b/tests/test_stokes.py
@@ -1,5 +1,17 @@
 import numpy as np
 from waveorder import stokes
+import pytest
+
+
+def test_A_matrix():
+    A5 = stokes.A_matrix(0.1)
+    assert A5.shape == (5, 4)
+
+    A4 = stokes.A_matrix(0.1, scheme="4-State")
+    assert A4.shape == (4, 4)
+
+    with pytest.raises(KeyError):
+        Ax = stokes.A_matrix(0.1, scheme="3-State")
 
 
 def test_s12_to_ori():
@@ -30,3 +42,17 @@ def test_stokes_recon():
 
                     for i in range(4):
                         assert ADR[i] - ADR1[i] < 1e-8
+
+
+def test_AR_mueller_from_CPL_projection():
+    # Check thank inv(M) == M.T, (only true when
+
+    M = stokes.AR_mueller_from_CPL_projection(
+        1, 1 / np.sqrt(3), 1 / np.sqrt(3), 1 / np.sqrt(3)
+    )
+    assert np.max(np.linalg.inv(M) - M.T) < 1e-8
+
+    M2 = stokes.AR_mueller_from_CPL_projection(
+        1, 1 / np.sqrt(2), 1 / np.sqrt(2), 0
+    )
+    assert np.max(np.linalg.inv(M2) - M2.T) < 1e-8
--- a/tests/test_stokes.py
+++ b/tests/test_stokes.py
@@ -1,6 +1,7 @@
 import numpy as np
 from waveorder import stokes
 import pytest
+import numpy.testing as npt
 
 
 def test_A_matrix():
@@ -10,7 +11,7 @@ def test_A_matrix():
     A4 = stokes.A_matrix(0.1, scheme="4-State")
     assert A4.shape == (4, 4)
 
-    assert np.max(stokes.A_matrix(0) - stokes.A_matrix(1)) < 1e-8
+    npt.assert_almost_equal(stokes.A_matrix(0), stokes.A_matrix(1))
 
     with pytest.raises(ValueError):
         Ax = stokes.A_matrix(0.1, scheme="3-State")
--- a/tests/test_stokes.py
+++ b/tests/test_stokes.py
@@ -1,0 +1,32 @@
+import numpy as np
+from waveorder import stokes
+
+
+def test_s12_to_ori():
+    for ori in np.linspace(0, np.pi, 25, endpoint=False):
+        ori1 = stokes._s12_to_ori(np.sin(2 * ori), -np.cos(2 * ori))
+        assert ori - ori1 < 1e-8
+
+
+def test_stokes_recon():
+
+    # NOTE: skip retardance = 0 and dop = 0 because orientation is not defined
+    for ret in np.arange(1e-3, 1, 0.1):  # fractions of a wave
+        for ori in np.arange(0, np.pi, np.pi / 10):  # radians
+            for tra in [0.1, 10]:
+
+                # Test attenuating retarder (AR) functions
+                AR = (ret, ori, tra)
+                s012 = stokes.s012_CPL_after_AR(*AR)
+                AR1 = stokes.inverse_s012_CPL_after_AR(*s012)
+                for i in range(3):
+                    assert AR[i] - AR1[i] < 1e-8
+
+                # Test attenuating depolarizing retarder (ADR) functions
+                for dop in np.arange(1e-3, 1, 0.1):
+                    ADR = (ret, ori, tra, dop)
+                    s0123 = stokes.s0123_CPL_after_ADR(*ADR)
+                    ADR1 = stokes.inverse_s0123_CPL_after_ADR(*s0123)
+
+                    for i in range(4):
+                        assert ADR[i] - ADR1[i] < 1e-8
--- a/tests/test_stokes.py
+++ b/tests/test_stokes.py
@@ -73,3 +73,19 @@ def test_AR_mueller_from_CPL_projection():
         1, 1 / np.sqrt(2), 1 / np.sqrt(2), 0
     )
     assert np.max(np.linalg.inv(M2) - M2.T) < 1e-8
+
+
+def test_mmul():
+
+    M = np.ones((3, 2, 1))
+    x = np.ones((2, 1))
+
+    y = stokes.mmul(M, x)  # should pass
+
+    with pytest.raises(ValueError):
+        M2 = np.ones((3, 4, 1))
+        y2 = stokes.mmul(M2, x)
+
+    with pytest.raises(ValueError):
+        M3 = np.ones((3, 2, 1, 2))
+        y3 = stokes.mmul(M3, x)
--- a/waveorder/stokes.py
+++ b/waveorder/stokes.py
@@ -102,9 +102,7 @@ def S2I_matrix(swing, scheme="5-State"):
             ]
         )
     else:
-        raise ValueError(
-            f"{scheme} is not implemented, use 4-State or 5-State"
-        )
+        raise ValueError(f"{scheme} is not implemented, use 4-State or 5-State")
     return S2I
 
 
@@ -314,7 +312,7 @@ def mueller_from_stokes(
     s3,
     input="CPL",
     model="AR",
-    direction="forward",
+    direction="inverse",
 ):
     """
     When light with input polarization state (default = circularly polarized
--- a/waveorder/stokes.py
+++ b/waveorder/stokes.py
@@ -1,9 +1,73 @@
 import numpy as np
 
+"""
+This module collects Stokes- and Mueller-related calculations. 
+
+The functions are organized into three groups:
+
+1) A forward function group: 
+
+A = A_matrix(swing, scheme="5-State")
+s0, s1, s3, s3 = s0123_CPL_after_ADR(ret, ori, tra, dop)
+s0, s1, s2 = s012_CPL_after_AR(ret, ori, dop)
+
+2) An inverse function group:
+
+inverse_s0123_CPL_after_ADR(s0, s1, s2, s3)
+inverse s012_CPL_after_AR(s0, s1, s2)
+AR_mueller_from_CPL_projection(s0, s1, s2, s3)
+
+3) A convenience function group:
+
+M = inv_AR_mueller_from_CPL_projection(s0, s1, s2, s3)
+y = matmul(A, x)
+
+----------
+
+Usage: 
+
+All functions (except A_matrix) are intended to be used with ND-arrays with
+Stokes- or Mueller-indices as the first axes. 
+
+For example, the following usage modes of s0123_CPL_after_ADR are valid:
+
+>>> s0123_CPL_after_ADR(1, 1, 1, 1)
+
+>>> ret = np.ones((2,3,4,5))
+>>> ori = np.ones((2,3,4,5))
+>>> tra = np.ones((2,3,4,5))
+>>> dop = np.ones((2,3,4,5))
+>>> s0123_CPL_after_ADR(ret, ori, tra, dop)
+
+>>> ADR_params = np.ones((4,2,3,4,5)) # first axis contains the Stokes indices
+>>> s0123_CPL_after_ADR(*ADR_params) # * expands along the first axis
+
+"""
+
+
 # Forward function group
 
 
 def A_matrix(swing, scheme="5-State"):
+    """
+    Calculate the polarimeter system matrix for a swing and calibration scheme.
+
+    Parameters
+    ----------
+    swing : float
+        Result is periodic on the integers, e.g. A_matrix(0.1) = A_matrix(1.1)
+    scheme : str, "4-State" or "5-State"
+        Corresponds to the calibration scheme used to acquire data,
+        by default "5-State"
+
+    Returns
+    -------
+    ArrayLike
+        Returns different shapes depending on the scheme
+
+        A.shape = (5, 4) for scheme = "5-State"
+        A.shape = (4, 4) for scheme = "4-state"
+    """
     chi = 2 * np.pi * swing
     if scheme == "5-State":
         A = np.array(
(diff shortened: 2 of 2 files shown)


mattersoflight:
> @ziw-liu your comments on a gpu switch, and on documentation+testing practice is very welcome. I wasn't sure if I should use type annotations & numpy-style docstrings? I stayed with only docstrings for now.

Re: GPU implementation, the API between pytorch and numpy seems quite consistent i.e., `object.operation` runs on CPU if object is a numpy array and on GPU if `object` is a torch tensor on GPU, e.g. [pinv](https://pytorch.org/docs/stable/generated/torch.linalg.pinv.html). If we are going to replace the GPU library, I'd first consider pytorch. Can you do a census of numpy methods you use and see if the same methods are available for torch tensor?

>@mattersoflight I'm particularly interested in your thoughts on naming. For example, inverse_s0123_CPL_after_ADR doesn't roll off the tongue like the earlier Polarization_recon, but I think it's important to be very specific at this level. Later layers of abstraction can reintroduce more abstract names likes Polarization_recon if we think they're helpful.

If we find that all the assumptions about polarization properties of the forward model can be covered by two properties: a) instrument matrix, and b) sample model, we can use a generic name and specify the assumptions via arguments. I'll think more about this.

mattersoflight:
`fft` is not relevant for this PR, but for deconvolution operations later. This benchmark reports that pytorch fft works almost as fast as cupy: https://thomasaarholt.github.io/fftspeedtest/fftspeedtest.html. pytorch is accelerated on M1, but cupy will require a nvidia gpu.

ziw-liu:
> Can you do a census of numpy methods you use and see if the same methods are available for torch tensor?

I wouldn't be too concerned about NumPy methods. However SciPy signal processing API may have a much lower coverage. Will have to check in detail.

ziw-liu:
Using torch is an interesting idea, in that it is 'accelerated' for CPUs too, so *in theory* the same code can work for CPU and GPU. However in addition to API coverage, lack of optimization/more overhead can be [potential issues](https://discuss.pytorch.org/t/torch-is-slow-compared-to-numpy/117502).

ziw-liu:
> I wasn't sure if I should use type annotations & numpy-style docstrings? I stayed with only docstrings for now.

We don't currently have type checking infra set up, so type hints serves mainly 2 purpose:

1. Help devs as well as users call the API in code elsewhere, because autocompletion and other in-editor static analysis works better.
2. Help generate the docstring. I personally use tools that populate the type info in the docstring automatically from docstrings.

I like to write type hints because it helps me code faster (e.g. I get syntax-highlighted and linted types that's copied over so less typos in the docstring type field). But as long as the code is consistent in style and well-documented I think it's all fine.

waveorder/stokes.py
```diff
@@ -0,0 +1,338 @@
...
+AR_mueller_from_CPL_projection(s0, s1, s2, s3)
+
+3) A convenience function group:
+
+M = inv_AR_mueller_from_CPL_projection(s0, s1, s2, s3)
+y = matmul(A, x)
+
+----------
+
+Usage:
```
ziw-liu, line 27: Minor styling issue:
Module docstring can also use NumPy-style and get auto-rendered in docs pages.
E.g. this section can be:
```docs
Examples
--------
Some text
    >>> print(1 + 3)  # some doctest code
    4
```

waveorder/stokes.py
```diff
@@ -0,0 +1,338 @@
...
+    ----------
+    s1, s2: ArrayLike, float
+        Stokes parameters
+
+    Returns
+    -------
+    ArrayLike, float
+        Slow-axis orientation with 0 <= ori < pi.
+    """
+    return (np.arctan2(s1, -s2) % (2 * np.pi)) / 2
```
talonchandler, line 192: @Soorya19Pradeep asked about consistency of this convention with the existing `waveorder` implementation. 

**My convention here is not consistent with the existing `waveorder_reconstructor` implementation**, but it is consistent with the  e-life paper, with the wikipedia article on retarders, and (most importantly) self consistent within this module. 

I can change this convention to be backwards compatible, but we will lose consistency with the e-life paper and (less importantly) it will require changes across this module to maintain self consistency. 

We have already placed some of the burden of orientation conventions on the user now that we have the "90 degree shift" checkbox in recOrder, so I'm leaning towards sticking with the convention implemented here. But I'm interested in hearing your thoughts @mattersoflight.

waveorder/stokes.py
ziw-liu: And they should also be rendered in supporting editors when imported. For example in VS code I can see the docstring of the NumPy module imported in this file.

waveorder/stokes.py
```diff
@@ -0,0 +1,338 @@
...
+
+def A_matrix(swing, scheme="5-State"):
+    """
+    Calculate the polarimeter system matrix for a swing and calibration scheme.
+
+    Parameters
+    ----------
+    swing : float
+        Result is periodic on the integers, e.g. A_matrix(0.1) = A_matrix(1.1)
+    scheme : str, "4-State" or "5-State"
```
ziw-liu, line 59: Another syntactical benefit of type hints is that `str, "4-State" or "5-State"` can be `Literal["4-State", "5-State"]` instead.

waveorder/stokes.py
talonchandler: Cool! Thanks for showing me this. Fixed.

waveorder/stokes.py
```diff
@@ -0,0 +1,338 @@
...
+                [
+                    1,
+                    -0.5 * np.sin(chi),
+                    -np.sqrt(3) * np.cos(chi / 2) * np.sin(chi / 2),
+                    -np.cos(chi),
+                ],
+            ]
+        )
+    else:
+        raise KeyError(f"{scheme} is not implemented, use 4-State or 5-State")
```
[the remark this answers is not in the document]
talonchandler: Thanks!

waveorder/stokes.py
talonchandler: This is where I got confused. Should I:

1) Change `scheme : str, "4-State" or "5-State"` to `scheme : Literal["4-State", "5-State"]`. Can a type checker use the docstring?

2) Use type annotations in the function signature? Does this mean I'll need to maintain both the docstring and the type annotation, or is there a shortcut?

ziw-liu: > Can a type checker use the docstring?

Not that I'm aware of.

> Use type annotations in the function signature? Does this mean I'll need to maintain both the docstring and the type annotation, or is there a shortcut?

In creation time docstrings can be polulated from the signature, but for the plugin I'm using they have to be manually synced in later editing. There might be other tools out there though.

Soorya19Pradeep:
@talonchandler, the cell membrane signal from the new orientation image computed with the additional background correction definitely has more contrast and is more continuous signal compared to the earlier version with just measured background correction. Thank you!

mattersoflight:
@talonchandler 

>@Soorya19Pradeep asked about the consistency of this convention with the existing waveorder implementation.

Could you clarify which convention we are discussing here:  convention for what is called right vs left circularly polarized light, or convention for axes of orientation, or may be both?

mattersoflight:
@ziw-liu , @talonchandler 
>Using torch is an interesting idea, in that it is 'accelerated' for CPUs too, so in theory the same code can work for CPU and GPU. 
My thought was to keep using numpy for CPU, and use torch instead of cupy for GPU. There can be code branches depending on whether you use CPU or GPU, but only when matrices (tensors) are instantiated.

Soorya19Pradeep:
> @talonchandler
> 
> > @Soorya19Pradeep asked about the consistency of this convention with the existing waveorder implementation.
> 
> Could you clarify which convention we are discussing here: convention for what is called right vs left circularly polarized light, or convention for axes of orientation, or may be both?

@mattersoflight, I am trying to understand how to read the orientation measurement of cell membrane, if it makes physical sense. The value of orientation changes with new implementation and further background correction, so I was curious. 
This is from orientation information from a cell from earlier version with measured background correction, colorbar range of values (-0.87,+0.87)
image
This is from the new version with just measured background correction, range (+0.87,+2.5). I realized the information here is same, just inverted and offset by 90 degrees.
image
After the extra background correction the range changes and more information is visible, range (0,+3.14).
image

talonchandler:
@mattersoflight 
> Could you clarify which convention we are discussing here: convention for what is called right vs left circularly polarized light, or convention for axes of orientation, or may be both?

Good question...I think we should discuss both. 

I can think of two paths to take here:

- **Decouple the reconstruction from the orientation convention.** We can give ourselves some flexibility in RCP vs. LCP, relative orientations of the camera wrt the polarizers, axis convention etc., then give the `recOrder` user (and the scripts) enough "knobs" to fix any deviation from convention. For example, we currently have one knob in recOrder that rotates by 90 degrees. To set the knobs, image a known sample (ideally a Kazansky target, but we can point our users to a more available alternative), twiddle the knobs until your colors match your expectations, then keep those knobs as default reconstruction parameters. **This is effectively what we're doing now, and I think it's workable.** 

- **Couple the reconstruction and orientation convention.** We can choose to be strict with our conventions: make the user specify RCP vs. LCP, the camera orientation, axis convention etc., then use those parameters as inputs to the reconstruction code. This will lead to the same results as above, but requires more diligence from `recOrder` user. In practice, I expect that this approach will result in the same approach as above---fiddle with these (physically motivated) knobs until you match your expectations of a known sample.

talonchandler:
Thanks for the comparison @Soorya19Pradeep. Very helpful. 

> I realized the information here is same, just inverted and offset by 90 degrees.

These are the types of knobs that we might provide in the "decoupling" approach: one checkbox/function for "invert orientation" and one for "rotate by 90 degrees".

mattersoflight:
> Decouple the reconstruction from the orientation convention. We can give ourselves some flexibility in RCP vs. LCP, relative orientations of the camera wrt the polarizers, axis convention etc., then give the recOrder user (and the scripts) enough "knobs" to fix any deviation from convention. For example, we currently have one knob in recOrder that rotates by 90 degrees. To set the knobs, image a known sample (ideally a Kazansky target, but we can point our users to a more available alternative), twiddle the knobs until your colors match your expectations, then keep those knobs as default reconstruction parameters. This is effectively what we're doing now, and I think it's workable.

This is the most general approach for any light path. I think two knobs suffice to register the angular coordinate system in data with the angular coordinate system on stage: one flip and one rotation.

mattersoflight:
> After the extra background correction the range changes and more information is visible, range (0,+3.14).
Thanks @Soorya19Pradeep for examples. It is great that you are taking a close look at FOVs. I suggest finding some patches where you see strong cortical actin bundles. If the background correction in this (arguably challenging) case is accurate, you'd see that orientation is parallel to the actin bundle. Once you have reconstructed retardance and orientation, you can call [`waveorder.visual.plotVectorField` ](https://github.com/mehta-lab/waveorder/blob/4b3b13364f313f752e23dde8bf9cf2080367acb4/waveorder/visual.py#LL995C18-L995C18) to visualize the orientation.

tests/test_stokes.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def test_A_matrix():
+    A5 = stokes.A_matrix(0.1)
+    assert A5.shape == (5, 4)
+
+    A4 = stokes.A_matrix(0.1, scheme="4-State")
+    assert A4.shape == (4, 4)
+
+    assert np.max(stokes.A_matrix(0) - stokes.A_matrix(1)) < 1e-8
```
ziw-liu, line 13: Does this line try to achieve [this function](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_almost_equal.html)?

waveorder/stokes.py
```diff
@@ -0,0 +1,344 @@
+"""
+Overview
+--------
+
+This module collects Stokes- and Mueller-related calculations. 
+
+The functions are organized into three groups:
```
ziw-liu, line 7: Not critical:
For now the forward/inverse/convenience are separated by comments in the source file. How about arranging them in a hierarchy so that from user side it's clear what function group it is:

```py
from waveorder import stokes

stokes.forward.A_matrix(...)

stokes.inverse.s0123_CPL_after_ADR(...)

stokes.utils.mmul(...)
```

waveorder/stokes.py
```diff
@@ -0,0 +1,344 @@
...
+    ----------
+    s0, s1, s2, s3 : NDArray, identical shapes
+        Stokes parameters
+
+    Returns
+    -------
+    NDArray, float, M.shape = (4, 4,) + s0.shape
+        Mueller matrix
+    """
+    M = np.zeros((4, 4) + np.array(s0).shape)
```
ziw-liu, line 276: Why is `np.array()` needed if `s0` is already an `NDArray` object?

waveorder/stokes.py
```diff
@@ -0,0 +1,344 @@
...
+    complete Mueller matrix of the AR.
+
+    Parameters
+    ----------
+    s0, s1, s2, s3 : NDArray, identical shapes
+        Stokes parameters
+
+    Returns
+    -------
+    NDArray, float, M.shape = (4, 4,) + s0.shape
```
ziw-liu, line 273: Can $s_{i}$ be 3D for this function? If not I suggest changing this to:

```
Returns
-------
NDArray[np.floating]
    Mueller matrix, shape is (4, 4, Y, X), where (Y, X) is the shape of `s0`.
```

Also note that if Python `float` (usually `float64`) is not strictly needed, [a more generic type](https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.floating) can be used.

talonchandler: s_i can be 1D, 2D, or 3D, but most commonly 1D or 2D. I think I'll leave it as is. 

Removed `float` since I don't think it's necessary to specify.

talonchandler: I'm going to reinstate `float` in the docs since it's convenient to have the option to give an ndarray or a float. I ran into this use case myself during testing.

waveorder/stokes.py
talonchandler: Because it's convenient to be able to provide an `NDArray` or a raw `float` (as a proxy for a 0D `NDArray`). 

I'm open to pushback if this is a bad idea.

waveorder/stokes.py
talonchandler: I'm very open to this. Let's discuss more tomorrow morning.

tests/test_stokes.py
talonchandler: Great call, thanks!

waveorder/stokes.py
ziw-liu: Sorry I misunderstood the typing again... didn't realize it was a union, not a member type.

waveorder/stokes.py
ziw-liu: Then maybe annotate here to?
https://github.com/mehta-lab/waveorder/blob/56eb82744e7c0cde712419b9cc4d3bae45cbbc83/waveorder/stokes.py#L268-L269

tests/test_stokes.py
ziw-liu: https://github.com/mehta-lab/waveorder/pull/110/commits/56eb82744e7c0cde712419b9cc4d3bae45cbbc83 changed this line, can you also change other occurrences, such as:
https://github.com/mehta-lab/waveorder/blob/56eb82744e7c0cde712419b9cc4d3bae45cbbc83/tests/test_stokes.py#L38
https://github.com/mehta-lab/waveorder/blob/56eb82744e7c0cde712419b9cc4d3bae45cbbc83/tests/test_stokes.py#L71-L72
so that they are consistent?

waveorder/stokes.py
```diff
@@ -0,0 +1,344 @@
...
+Overview
+--------
+
+This module collects Stokes- and Mueller-related calculations. 
+
+The functions are organized into three groups:
+
+1) A forward function group: 
+
+A = A_matrix(swing, scheme="5-State")
```
mattersoflight, line 11: We should use a more descriptive name here: `A_matrix` should be called `Instrument_matrix`, `Sys_matrix`, `I2Stokes_matrix`.

waveorder/stokes.py
```diff
@@ -0,0 +1,344 @@
...
+s0, s1, s3, s3 = s0123_CPL_after_ADR(ret, ori, tra, dop)
+s0, s1, s2 = s012_CPL_after_AR(ret, ori, dop)
```
mattersoflight, line 13: These names can read like a jargon. How about `stokes_ADR_CPL` and `stokes_AR_CPL`. Writing down the mnemonics for components in the order in which their Mueller matrices multiply can make it easier to discover the methods if we add more transformations.

waveorder/stokes.py
```diff
@@ -0,0 +1,344 @@
...
+
+2) An inverse function group:
+
+ret, ori, tra, dop = inverse_s0123_CPL_after_ADR(s0, s1, s2, s3)
+ret, ori, tra = inverse_s012_CPL_after_AR(s0, s1, s2)
```
mattersoflight, line 18: I suggest:
ret, ori, tra, dop = inverse_ADR(s0,s1,s2,s3,input='CPL')
ret, ori, tra = inverese_AR(s0,s1,s2,input='CPL')

waveorder/stokes.py
```diff
@@ -0,0 +1,344 @@
...
+M = AR_mueller_from_CPL_projection(s0, s1, s2, s3)
```
mattersoflight, line 19: I suggest:
`M = mueller_from_projection(s0,s1,s2,s3, model='AR', input='CPL')`

waveorder/stokes.py
```diff
@@ -0,0 +1,344 @@
...
+
+3) A convenience function group:
+
+M = inv_AR_mueller_from_CPL_projection(s0, s1, s2, s3)
```
mattersoflight, line 23: I suggest:
```
M = mueller_from_projection(s0,s1,s2,s3, model='AR', direction='inverse', input='CPL')
```

mattersoflight (commented):
Overall, this PR is a great step forward towards unifying the polarization algebra in our codebase.


The module currently consists of many small methods. They can be unified by specifying some of the assumptions by parameters. 

I've made specific suggestions below. With that, the script can look like this.

```
# Calculate A and A_inv
A = stokes.instrument_matrix(swing=0.1, scheme="5-State")
A_inv = np.linalg.pinv(A)

# Apply A_inv to background and sample data
S_bg = stokes.mmul(A_inv, cyx_bg_data)
S_sm = stokes.mmul(A_inv, czyx_data)

# Calculate background correction matrix from S_bg
M_inv = stokes.mueller_from_projection(*S_bg. model='AR', direction='inverse', input='CPL')

# Apply background correction to sample data
bg_corr_S_sm = stokes.mmul(M_inv, S_sm)

# Reconstruct parameters
ret, ori, tra, dop = stokes.inverse_ADR(*bg_corr_S_sm, input = 'CPL')

```

Additional thoughts:
* I think that estimate_ADR is a reasonable substitute for inverse_ADR.
* I went back and forth about how to shape the ND arrays s0, s1, s2, s3. On one hand, they can be just one argument with a dimension serving as Stokes. But on the other hand, it may complicate implementation and require expensive transposes. This requires quick discussion in light of https://github.com/mehta-lab/recOrder/pull/322

I can run the example script once the code path is refactored.

waveorder/stokes.py
talonchandler: I'm considering changing my docs from `NDArray` back to `array_like` since that's what `numpy` does.

I'm aiming for behavior like most numpy functions. If you pass in "raw numbers", it'll return a raw number (`type(np.arctan2(1.0,1.0))` gives `np.float64`), and if you pass in an array it'll return an array (`type(np.arctan2(np.array([1.0,2.0]),np.array([1.0,2.0])))` gives `np.ndarray`).

talonchandler: I think documenting with the `array_like` type implies these cases, since it's how `numpy` is documented.

ziw-liu: `ArrayLike` is [a `Union` of many different things](https://github.com/numpy/numpy/blob/main/numpy/_typing/_array_like.py). If the purpose is to accept `float`s and `NDArray`s, but not Dask arrays (implements array protocol) and lists of lists (nested sequence), `Union[np.floating, NDArray]` is more accurate.

talonchandler: These functions are thin wrappers for `numpy` functions that are documented as accepting `array_like` types, so I don't understand why I can't also document these functions as accepting `array_like` types. They should work on Dask arrays and lists of lists if the underlying `numpy` arrays also accept these. I'm only using `.shape` and `np` functions that accept `array_like`s.

waveorder/stokes.py
talonchandler: I just renamed `A_matrix` to `I2S_matrix` and added a new `S2I_matrix` function. 

I'm open to more iteration on these names.

waveorder/stokes.py
talonchandler: I think adding `input="CPL"` is a big improvement!

How do these sound?
`stokes_after_ADR(ret,ori,tra,dop, input='CPL')`
`stokes012_after_AR(ret,ori,tra, input='CPL')`

I think it's important to make it clear that these are stokes vectors _after_ an ADR/AR, and I think if we don't return all 4 stokes params then the function name should make that clear. 

> Reading the mnemonics for components in the order in which vectors and matrices multiply is more intuitive. The input can also be a Stokes vector when one wants to use the methods for another purpose, e.g., model non-ideal illumination.

I'm not sure I understand this. The matched `inverse` functions below take Stokes vectors as input...is this what you have in mind?

waveorder/stokes.py
talonchandler: Similar to above...how do these sound:
```
ret, ori, tra, dop = inverse_stokes_after_ADR(s0, s1, s2, s3, input="CPL")
ret, ori, tra = inverse_stokes012_after_AR(s0, s1, s2, input="CPL")
```

`inverse_ADR` alone might make me think that I'm going to get a Mueller matrix.

waveorder/stokes.py
talonchandler: Much better!

Thinking more, I also don't think `projection` is the right word here. I think:

`M = mueller_from_stokes(s0, s1, s2, s3, model="AR", direction="forward", input="CPL")`

or possibly `mueller_from_measured_stokes` are even clearer.

waveorder/stokes.py
talonchandler: Further down you suggest `estimate_ADR`, and I do like the word `estimate` here. 

I'm now favoring 
```
ret, ori, tra, dop = estimate_ADR_from_stokes(s0, s1, s2, s3, input="CPL")
ret, ori, tra = estimate_AR_from_stokes012(s0, s1, s2, input="CPL")
```

talonchandler:
I've just completed the renaming/refactoring. @mattersoflight this is ready for your re-review. 

The latest API (at this level) looks like:
```
# Calculate I2S
I2S = stokes.I2S_matrix(swing, scheme="5-State")

# Calculate Stokes vectors
S_sm = stokes.mmul(I2S, czyx_data)  # Apply I2S to sample data
S_bg_meas = stokes.mmul(I2S, cyx_bg_data)  # Apply I2S to measured background

# Calculate background correction matrix from S_bg
M_bg_inv = stokes.mueller_from_stokes(*S_bg_meas)

# Apply measured background correction to sample data
bg_corr_S_sm = stokes.mmul(M_bg_inv, S_sm)

# Recover ADR parameters
ret, ori, tra, dop = stokes.estimate_ADR_from_stokes(*bg_corr_S_sm)
```

I've also spent some time characterizing the old vs. new algorithms.

**Soorya's cells - retardance on y axis - measured bkg correction only**
Screenshot 2023-03-06 at 5 53 03 PM
At most 2% difference in retardance.

**Kazansky target - retardance on y axis - measured bkg correction only**
Screenshot 2023-03-06 at 5 31 36 PM
At most 1% difference in retardance.

**Kazansky target - orientation on y axis - measured bkg correction only**
Screenshot 2023-03-06 at 5 35 18 PM
At most 2% difference in non-background regions when the different orientation convention is accounted for. This main difference here is from a difference in orientation conventions which we'll be handling with two user-facing switches as discussed above. 

**Timing**
Current performance bottleneck is the pre-calculation of `mueller_from_stokes` from the background stokes vectors, which can be further optimized (I expect factors of 2-4x). For now, here are a couple benchmarks: 

1 x 2048 x 2048: 
old algorithm: 1.0 s 
new algorithm: 15.4 s 

8 x 2048 x 2048: 
old algorithm: 17.8 s 
new algorithm: 19.6 s 

**Example comparison script (generates Kaz target comparison above)**

<details>
  <summary>Full example script (click to expand):</summary>

```
import numpy as np
from waveorder import stokes
from recOrder.io.utils import load_bg
from recOrder.io.metadata_reader import MetadataReader
from recOrder.compute.reconstructions import (
    initialize_reconstructor,
    reconstruct_qlipp_stokes,
    reconstruct_qlipp_birefringence,
)
from iohub.reader import imread
from iohub.ngff import open_ome_zarr
import napari

# Set paths
base_path = "/hpc/projects/compmicro/rawdata/hummingbird/Talon/2023_02_08_kaz/"
data_subpath = "kaz-raw_recOrderPluginSnap_0/kaz-raw_RawPolDataSnap.zarr"
bg_subpath = "BG"
cal_subpath = "calibration_metadata.txt"

# Read data
reader = imread(base_path + data_subpath)
T, C, Z, Y, X = reader.shape
czyx_data = reader.get_array(position=0)[0, ...]

# Read background data
cyx_bg_data = load_bg(base_path + bg_subpath, height=Y, width=X)

# Read calibration metadata
md = MetadataReader(base_path + cal_subpath)

def new_bg_correction(czyx_data, cyx_bg_data, swing, scheme):
    # Calculate I2S
    I2S = stokes.I2S_matrix(md.Swing, scheme=md.get_calibration_scheme())

    # Calculate Stokes vectors
    S_sm = stokes.mmul(I2S, czyx_data)  # Apply I2S to sample data
    S_bg_meas = stokes.mmul(I2S, cyx_bg_data)  # Apply I2S to measured background

    # Calculate background correction matrix from S_bg
    M_bg_inv = stokes.mueller_from_stokes(*S_bg_meas)

    # Apply measured background correction to sample data
    bg_corr_S_sm = stokes.mmul(M_bg_inv, S_sm)

    # Recover ADR parameters
    ret, ori, tra, dop = stokes.estimate_ADR_from_stokes(*bg_corr_S_sm)

    ret = ret / (2 * np.pi) * 532

    return ret, ori, tra, dop

def old_bg_correction(czyx_data, cyx_bg_data, swing, scheme):
    reconstructor_args = {
        "image_dim": (Y, X),
        "n_slices": 1,  # number of slices in z-stack
        "wavelength_nm": 532,
        "swing": swing,
        "calibration_scheme": scheme,  # "4-State" or "5-State"
        "bg_correction": "global",
    }
    reconstructor = initialize_reconstructor(
        pipeline="birefringence", **reconstructor_args
    )
    # Reconstruct background Stokes
    bg_stokes = reconstruct_qlipp_stokes(cyx_bg_data, reconstructor)

    # Reconstruct data Stokes w/ background correction
    stokes = reconstruct_qlipp_stokes(czyx_data, reconstructor, bg_stokes)

    birefringence = reconstruct_qlipp_birefringence(stokes, reconstructor)
    birefringence[0] = (
        birefringence[0] / (2 * np.pi) * reconstructor_args["wavelength_nm"]
    )
    return birefringence

oldADR = old_bg_correction(czyx_data, cyx_bg_data, md.Swing, md.Calibration_scheme)
newADR = new_bg_correction(czyx_data, cyx_bg_data, md.Swing, md.Calibration_scheme)


v = napari.Viewer()
v.add_image(oldADR[..., 890:1220, 790:1370], name="old")
v.add_image(np.stack(newADR)[..., 890:1220, 790:1370], "new")
import pdb; pdb.set_trace()
```
</details>

waveorder/stokes.py
```diff
@@ -0,0 +1,406 @@
...
+            s0, s1, s2, s3, input=input, model=model, direction="forward"
+        )
+        M_flip = np.moveaxis(M, (0, 1), (-2, -1))
+        M_inv_flip = np.linalg.inv(M_flip)  # applied over the last two axes
+        M_inv = np.moveaxis(M_inv_flip, (-2, -1), (0, 1))
+        return M_inv
+
+
+def mmul(matrix, vector):
+    """Convenient matrix-multiply used for
```
ziw-liu, line 390: Naïve question: how is this function different from `np.matmul()`?

talonchandler: Very similar! But this function handles the dimension ordering more smoothly that `np.matmul`.

1) `np.matmul` acts on the last two dimensions of the arrays, while this function acts on the first two dimensions, which helps us reduce error- and copy-prone transposes in our CZYX-order arrays. 

2) `np.matmul` has tricky rules (i.e. rules that I've never completely grokked) about how it handles vectors and matrices. IME it's easiest to use if you make everything a matrix so that the last indices match `(n,k),(k,m)->(n,m)`, but this requires `np.newaxis` calls.

It's reasonably fair to compare this one-line einsum call to the `Stokes_recon` call here: 
https://github.com/mehta-lab/waveorder/blob/4b3b13364f313f752e23dde8bf9cf2080367acb4/waveorder/waveorder_reconstructor.py#L1745

When there's `move_axis` or `transpose` calls before and after a `matmul` or similar operation, you can almost always simplify with an `einsum`. 

Finally, there are (very reasonable) concerns about timing, and I just ran a quick comparison on these types of operations:

```
import numpy as np
import time

N = 1000

# Einsum
A = np.random.random((4, 4, N, N))
x = np.random.random((4, N, N))

t = time.time()
y1 = np.einsum('ijkl,jkl->ikl', A, x)
print(f"einsum: {time.time() - t:.4f}")

# matmul
At = A.transpose((2,3,0,1))
xt = x.transpose((1,2,0))[...,np.newaxis]

t = time.time()
y2 = np.matmul(At, xt)
print(f"matmul: {time.time() - t:.4f}")

assert np.max(y2[...,0].transpose((2,0,1)) - y1) < 1e-8
```
```
einsum: 0.0185
matmul: 0.0234
```
gives `einsum` a 20% edge over `matmul`, even without considering the cost of the transposes (very cheap if there are no copies, but very expensive if you do need to copy).

talonchandler:
I neglected to commit one small change: `stokes.mueller_from_stokes` should be `direction=inverse` by default since this is the most common usage mode (and the usage mode I showed in the snippet from last night).

mattersoflight:
thanks for the offline discussion. Looks great to me!

PR #114, opened 2023-04-16
Prepare for `waveorder==1.0.0` release

talonchandler:
I am grouping our remaining tasks to prepare for `1.0.0` into this PR. Please feel free to edit or suggest additional tasks.  

- [x]  remove `WaveorderReader`, `WaveorderWriter` and associated tests from `waveorder`
- [x] identify example scripts that we need to maintain - @mattersoflight can you help with this?
- [x] remove unmaintained examples, and ~transition the maintained example scripts to `iohub`~ and test
- [ ] merge this PR
- [ ] release `waveorder 1.0.0`

This PR needs to merge before `alg-dev` merges.


mattersoflight:
>  identify example scripts that we need to maintain - @mattersoflight can you help with this?

Hi @talonchandler,  here are the scripts we don't need to maintain in this repo:
* [multiProcessing](https://github.com/mehta-lab/waveorder/tree/main/examples/multiProcessing) can be moved to recOrder. 
* [Reader_Writer_Examples](https://github.com/mehta-lab/waveorder/tree/main/examples/Reader_Writer_Examples) can be deleted (since we now use iohub 🎉 ).
* We should maintain all the other forward simulation and reconstruction [example](https://github.com/mehta-lab/waveorder/tree/main/examples) as Python scripts. We cannot maintain corresponding jupyter notebooks. They can be replaced with time-stamped PDFs within `docs` that are updated with every release.  
* The maintained examples can be refactored as the code path evolves. These examples should also be extended with any scripts you use on regular basis. 
 
> Transition the maintained example scripts to iohub and test
Are you thinking of turning these examples via pytest? That's an interesting idea and will allow us to test refactors. We should discuss offline which of the tests should be run during CI and which should be run manually before release.

Once we settle on where the PDFs should go, I'll start to restructure the examples.

mattersoflight:
@edyoshikun can you help with moving the multiProcessing scripts to recOrder?

edyoshikun:
I think we should only have 1 example for the multiprocessing on the recOrder side since the codebase is the same. The current example from @talonchandler [here](https://github.com/mehta-lab/recOrder/blob/main/examples/parallel-reconstruct.py) is a great example and the slurm scripts can serve as compliment [slurm scripts PR](https://github.com/mehta-lab/recOrder/pull/337). Also from the testing and future development and refactoring perspective, we agree it is easier to mantain.

talonchandler:
This PR is ready for review. After discussing with @mattersoflight and looking closely at all of the examples, I partitioned the examples into two sets:
1. Maintained example scripts that "just work". I have updated scripts that demonstrate QLIPP, 3D phase, and PTI simulations and reconstructions, and I've added them to CI/CD so that we will know when they're failing. I am planning to maintain these scripts as we move through the refactor. All of these are `.py` files. 
2. `experimental_reconstructions` notebooks that are dependent on specific datasets, take a longer time to run, and include a lot of duplicated code. I am planning to keep these as a form of documentation, but I am not planning to add these to CI/CD or maintain them through the upcoming refactoring. 

@edyoshikun I agree that the multiprocessing example in `recOrder` is a cleaner version of the work here, so I removed these examples. 

I'm very open to suggestions. Does the folder structure makes sense (I'm hoping the new `examples/README.md` helps)? I haven't added any timestamped PDFs as you suggested, but I *think* that the notebooks in `experimental_reconstructions` serve the same purpose?

mattersoflight (dismissed):
Thanks for the reorganization. Looks great!

talonchandler:
@ziw-liu would you mind taking a quick look at this as well? (Also I accidentally "dismissed" the earlier review by fixing a small problem.)

@ziw-liu when this merges it will be ready for `waveorder==1.0.0` release.

examples/PTI_simulation/PTI_Simulation_Recon3D.py
```diff
@@ -499,9 +499,9 @@
 
 ### select slices to plot ###
 spacing = 4
-z_layer = 50
-x_layer = 100
-y_layer = 100
+z_layer = L//2
+x_layer = N//2
+y_layer = M//2
```
ziw-liu, line 504: Formatting: missing spaces.

examples/README.md
```diff
@@ -0,0 +1,12 @@
+This folder of examples consists of:
+
+1) A set of simulations that will run as is. Any of the files in these folders 
+can be run with `python <script-name>.py`, and these files are included in 
+`waveorder`'s automated testing.  
+- `/QLIPP_simulation/`
```
ziw-liu, line 6: Is this name supposed to be matching the directory name?

tests/test_examples.py
```diff
@@ -0,0 +1,25 @@
...
+        "3D_PODT_phase_simulation/3D_PODT_Phase_forward.py",
+        "3D_PODT_phase_simulation/3D_PODT_Phase_recon.py",
+        "PTI_simulation/PTI_Simulation_Forward_2D3D.py",
+        "PTI_simulation/PTI_Simulation_Recon2D.py",
+        "PTI_simulation/PTI_Simulation_Recon3D.py",
+    ]
+
+    for script in scripts:
+        path = os.path.join(os.getcwd(), "examples/", script)
+        completed_process = subprocess.run(["python", path])
```
ziw-liu, line 24: Another way is to have a main function for these scripts and call them in separate test cases. That way if a subset fails it's easier to debug. But that may impair readability by a little bit?

tests/test_examples.py
```diff
@@ -0,0 +1,25 @@
+import subprocess
+import os
+from pathlib import Path
+
+
+# Currently, these tests will run in the background if no screen is available
+# If you do have a screen available, plots will appear, and you'll need to
+# close them to make the tests pass.
+# TODO: see if we can make these run locally w/o showing matplotlib
+# @patch("matplotlib.pyplot.show")
```
ziw-liu, line 10: Does this work?

waveorder/visual.py
```diff
@@ -414,7 +407,7 @@ def parallel_4D_viewer(
         raise ValueError(
             "number of titles does not match with the number of channels"
         )
-    num_row = np.int(np.ceil(N_channel / num_col))
+    num_row = int(np.ceil(N_channel / num_col))
```
ziw-liu, line 410: This reminds me of NEP29: the advised support range for NumPy is now 1.21+

talonchandler: Thanks!

examples/PTI_simulation/PTI_Simulation_Recon3D.py
talonchandler: Good eye. I wonder if I have a weird setting with my vscode black formatter...I've been getting the odd glitch like this (also adding and removing empty lines).

examples/README.md
talonchandler: great eye.

tests/test_examples.py
talonchandler: Not yet...I spent ~15 mins trying a few things and found the commented decorator here: https://stackoverflow.com/questions/51429465/how-to-mock-pyplot-show-in-python-to-prevent-showing-plots

But it doesn't seem to work as I want it to, and I don't have the capacity right now to completely understand how these mocks+decorators work. 

It's a low priority, since this runs smoothly for now...but if you have experience here I'd love to learn.

ziw-liu: I suspect that this doesn't work because the in-memory patch is not propagated to the subprocesses.

ziw-liu: But we can deal with this later!

tests/test_examples.py
talonchandler: Thanks for the suggestion...I've cleaned this up a bit. 

FYI some of these scripts are dependent on the earlier scripts (the recons depend on the forward simulations), which is why I've grouped them into three groups.

ziw-liu (approved):

PR #117, opened 2023-04-20
Phase 2D + 3D refactor

talonchandler:
This PR makes the following changes to the "Phase2D" and "Phase3D" reconstructions methods:
- moves these code paths from `waveorder_reconstructor` into their own files in `/waveorder/models/phase2D_3D.py` and `waveorder/models/phase3D_3D.py`. 
- refactors both of these reconstruction paths into the following methods: `calculate_transfer_function`, `visualize_transfer_function` (for napari viewing), `apply_transfer_function` (very simple simulation),  and `apply_inverse_transfer_function`. These methods are in preparation for direct calls from the `recOrder` CLI. 
- performs all calculations with `torch`
- performs all calculations in CZYX order 
- adds tests to the key parts of `optics`, `utils`, and the new `examples`
- preserves the code paths for the PTI and `Birefringence_recon` (S1&S2 diffraction-aware reconstruction) in `waveorder_reconstructor`. I am now testing several example scripts to alert me to breaking changes. 

**I suggest starting your review by running the new examples scripts in `/examples/PODT_Phase_simulation/`**. This will demonstrate a simplified form of what I expect on the `recOrder` side: the TF is calculated (and optionally viewed in napari), saved to a file, then used to reconstruct the data. 

**Next, I suggest you take a look at the (currently under-documented but hopefully much clearer than before) function signatures for `calculate_transfer_function` and `apply_inverge_transfer_function` in `models/phase2D_3D.py` and `models/phase3D_3D.py`.** These function signatures have a nearly direct correspondence with `recOrder`'s GUI parameters.

**Finally, I'll request your thoughts on my `models` format that arose naturally during my refactor.** I've temporarily named the three models after our assumption about the sample (`phase2D`,`phase3D`, `planaranisoND`) and the data we acquire (`3D`, `polND`). To my eye, this seems to be an extensible way to split up the code since different assumptions about the object and/or different data will lead to different models (and code). I'm also very aware that this *could* be a good application for an abstract base class (especially since I expect to implement `calculate_transfer_function` and `apply_inverse_transfer_function` methods for `planaranisoND_polND`), but I've chosen to delay this for now in fear of premature abstraction.

 I'm happy to chat in person anytime tomorrow. My immediate focus is on the interface for the "Retardance + Orientation" pipeline which I expect to require about 3/4 of my day tomorrow. After that PR, I am planning to start working on the `recOrder` CLI side and link everything up.


examples/PODT_phase_simulation/phase2D_3D.py
```diff
@@ -0,0 +1,57 @@
+# 3D partially coherent optical diffraction tomography (ODT) simulation
+# J. M. Soto, J. A. Rodrigo, and T. Alieva, "Label-free quantitative
+# 3D tomographic imaging for partially coherent light microscopy," Opt. Express
+# 25, 15699-15712 (2017)
```
ziw-liu, line 4: For comment like this you can append `#noqa` to tell flake8 that this should not be linted.

waveorder/models/phase2D_3D.py
```diff
@@ -0,0 +1,99 @@
+import torch
+import numpy as np
+from waveorder import optics, util
+
+
+def calc_TF(YX_shape, YX_ps, Z_pos_list, lamb_ill, n_media, NA_ill, NA_obj):
```
ziw-liu, line 6: Generally function names, parameters, and local variables should be snake_case according to [PEP8](https://peps.python.org/pep-0008/#function-and-variable-names).

ziw-liu: Also I wouldn't be too worried about having longer function names. Using complete words improve readability and the code can be auto-completed anyway.

waveorder/models/phase2D_3D.py
```diff
@@ -0,0 +1,99 @@
...
+            ill_pupil, det_pupil * Hz_stack[z]
+        )
+
+    return Hu, Hp
+
+
+def visualize_TF(viewer, Hu, Hp, ZYX_scale):
+    # TODO: consider generalizing w/ phase3Dto3D.visualize_TF
+    arrays = [
+        (torch.imag(Hu).numpy(), "Im(H_u)"),
```
ziw-liu, line 30: In the extreme case (when the tensor is on a GPU and being differentiated), one would need `tensor.detach().cpu().numpy()` to get the NumPy array.

Since we won't compute gradients in the near term, `tensor.cpu().numpy()` is enough.

waveorder/models/phase2D_3D.py
```diff
@@ -0,0 +1,99 @@
...
+        (torch.real(Hu).numpy(), "Re(H_u)"),
+        (torch.imag(Hp).numpy(), "Im(H_p)"),
+        (torch.real(Hp).numpy(), "Re(H_p)"),
+    ]
+
+    for array in arrays:
+        lim = 0.5 * np.max(np.abs(array[0]))
+        viewer.add_image(
+            np.fft.ifftshift(array[0], axes=(1,2)),
```
ziw-liu, line 39: Should we do this with tensors?

examples/PODT_phase_simulation/phase2D_3D.py
talonchandler: Discussed in person, and there's no change needed here. Thanks.

waveorder/models/phase2D_3D.py
talonchandler: I found that giving a tensor to `viewer.add_image()` fails, so I found I needed to convert with `.numpy()`. Should I always use `.cpu().numpy()`, or is this issue small enough that we can handle it if we need it?

ziw-liu: I meant doing the FFT in torch and then converting to NumPy, so you don't have to do it many times upfront.

For now `cpu()` doesn't do anything, just to make it future-proof when we do support GPU.

waveorder/models/phase2D_3D.py
talonchandler: I've updated all of the names in the `phase3D_3D` and `phase2D_2D` code paths. I'll resolve this comment, but please feel free to continue to make suggestions.

waveorder/models/phase2D_3D.py
talonchandler: Fixed, thanks!

waveorder/models/phase2D_3D.py
talonchandler: Fixed.

talonchandler:
Thanks for the contributions and in-person chat @mattersoflight. I've made the following changes:

- renamed the models using the schema `<object-type>_<object-thickness>_<data-shape> `. We're starting with `isotropic_thin_3d`, `phase_thick_3d`, and `inplane_anisotropic_thin_pol3d`. 
- fixed `isotropic_thin_3d` so that it handle absorption, and fixed the example to simulate a thin object. 
- renamed from `2D/3D -> 2d/3d` following PEP8 and one of Keith's comments in the dragonfly repo.

talonchandler:
Merging into `alg-dev`...we can discuss more when I open a PR into `main`

PR #118, opened 2023-04-22
Prepare polarization algorithms for integration with `recOrder`

talonchandler:
This branch prepares the polarization algorithms in `stokes.py` to be used by `recOrder`. 

- new `inplane_anisotropic_thin_pol3d.py` model and example file with `calculate_transfer_function`, `apply_inverse_transfer_function` functions. These are quite thin wrappers on `stokes.py`, and they add cleaner background-correction strategies for `recOrder` to use directly
- revised `stokes.py` to use more verbose names: `ret` -> `retardance` etc. 

@mattersoflight @ziw-liu if you want to take a closer look you can start by running `/examples/models/inplane_anisotropic_thin_pol3d.py`. This is a very simple simulation that focusses on what `recOrder` needs---the simulation does not currently include any diffraction on the forward or inverse models. 

Over the weekend I will start integrating the new interface on the `recOrder` side. I will be doing integration work from this branch (and/or `alg-dev`) and a new branch (`cli-dev`) on the `recOrder` side so I'm not blocked before your review.


talonchandler:
Merging into `alg-dev`.

PR #121, opened 2023-06-07
`alg-dev` branch merge to `main`

talonchandler:
The `alg-dev` branch has been the de-facto `main` branch for the last ~2 months, so I think it's a good idea for us to merge. 

The major changes (discussed in their individual PRs to `alg-dev`) are:
- new Stokes calculus (well tested with `recOrder`)
- new `models`-based reconstructions with precalculation of transfer functions
- cleaned and tested `examples` folder


ziw-liu (dismissed):

talonchandler:
I'm now planning to solve #122 before merging this.

mattersoflight:
Perhaps we should use the word `orientation` or `oriented` instead of `anisotropy`. See [here](https://github.com/mehta-lab/waveorder/issues/122#issuecomment-1587568830) for why doing so makes sense for fluorescence orientation measurement.

waveorder/models/inplane_anisotropic_thin_pol3d.py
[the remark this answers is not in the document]
talonchandler: Great, yes I agree `thin` should be `thick`. Thanks. 

I agree that `anisotropy` is heavily overloaded and worth avoiding. I lean towards `oriented` since it matches with `isotropic` and `fluorescent`.

So I think I'm settling towards `inplane_oriented_thick_pol3d`. Thanks for the discussion.

setup.cfg
```diff
@@ -46,11 +46,11 @@ install_requires =
     zarr>=2.6.1
     natsort>=7.1.1
     pycromanager>=0.13.2
+    torch
```
edyoshikun, line 49: Not sure if we should pin a version(?)

talonchandler: I don't think we need to yet. Unless you've had a fresh install fail for versioning reasons?

waveorder/models/isotropic_thin_3d.py
```diff
@@ -0,0 +1,217 @@
...
+
+    yx_absorption = 0.99 * sphere[1]
+
+    return yx_absorption, yx_phase
+
+
+def calculate_transfer_function(
+    yx_shape,
+    yx_pixel_size,
+    z_position_list,
```
ziw-liu, line 36: I didn't catch this in #131, but we need a better name and/or more documentation to indicate that this is a `torch.Tensor`, not a `list`.

ziw-liu: (Looking more into the history it used to be a `list`.)

setup.cfg
ziw-liu: There is no need to pin torch if we are only doing a fresh installation. But people may force it in some other environments where e.g. torch<1.8 is required, and that may bear minor differences from 2.0+ that we are testing against.

talonchandler: Sounds good. @ziw-liu would you suggest `torch>=2.0.0`? That's what I've been testing against, and I just added.

waveorder/models/isotropic_thin_3d.py
talonchandler: Great catch @ziw-liu, and your test case helped me sort this out.

Initially, I named it `_list` because I wanted it to be a `list`. But then your tests were failing and I "fixed" it by passing a tensor, which was a false fix.

Now I'm passing in a list in the test case, and I fixed the issue.

ziw-liu (approved):

PR #124, opened 2023-06-09
Add option for axial flip of `phase_thick_3d` transfer function

talonchandler:
This PR solves a common problem in "3D phase" reconstructions: you perform a reconstruction and see that the contrast is inverted. Should you flip the image? Flip the transfer function? Do these options add any computational expense?

This option lets you flip the transfer function without any additional computational expense or copying.  

Note: I have not added a similar option to `isotropic_thin_3d` because it takes a `z_position_list` directly, so the user can flip that list directly.

### DIFF
--- a/tests/models/test_phase_thick_3d.py
+++ b/tests/models/test_phase_thick_3d.py
@@ -1,7 +1,9 @@
+import pytest
 from waveorder.models import phase_thick_3d
 
 
-def test_calculate_transfer_function():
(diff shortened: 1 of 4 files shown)


talonchandler:
Sorry for the accidental inclusion of #123.

ziw-liu (commented):
Can we have a test case for this?

ziw-liu (approved):

talonchandler:
Thanks @ziw-liu!

PR #132, opened 2023-07-11
Add optional parameters `plot_path` and `threshold_FWHM` to `focus_from_transverse_band` 

### LINKED ISSUE #129: Focus-finding features
@ieivanov and I just discussed some desired features for the focus-finding algorithms:

- return a normalized "SNR" or "confidence" metric alongside the in-focus slice. We're concerned that low-SNR images may make the focus-finding fail and cause erratic stage motion. I need to think more and test exactly what metric we'll use, but @ieivanov and I agreed that we want the metric to be invariant to (1) the number of pixels, (2) the overall brightness of the sample, and (3) the size of the sample. 
- a `plot=True` option to save a small figure of the focus-finding run, likely mid-band power vs. slice, a dot for the peak, and an annotation for the confidence metric.

talonchandler:
**Summary**
This PR fixes #129 and adds:
- an optional `plot_path` option...passing a path string will save a diagnostic plot to that path (`.png`, `.svg`, `.pdf` etc)
- an optional `threshold_FWHM` option (default 0) will make the function return an in-focus slice if the mid-band power peak has a FWHM larger than the threshold...otherwise the function will return `None`. 

**Rationale:**
After my initial conversation with @ieivanov, I looked for a way to "normalize" the mid-band power metric so that we could try using an absolute threshold. I ended up convincing myself that finding an absolute threshold would be challenging. 

For example, consider an in-focus object that fills the FOV then the same object that fills a small part of the FOV (padded with zeros/noise). If we normalize by the number of pixels in the FOV then the mid-band power appears to change between these two objects, when we'd like to focus on them both equally well. I played with a few alternative normalizations, and ultimately decide to try a different approach. 

I ended up settling on the `threshold_FWHM` approach implemented here. The rationale is that noise will typically have peaks with smaller FWHM than real axially-extended in-focus objects, so we can reject "in-focus noise" by thresholding on the FWHM. Larger FWHM peaks are also associated with peaks that have tighter sampling (i.e. we have more confidence that the object is there) and that are away from the edges (a peak at the very edge has zero FWHM by definition in the current implementation). 

**Performance on real data**
Caption: four focus-finding diagnostic plots generated with real data (2023-07-03). Rows: two positions. Columns: whole FOV and a hand-selected background region filled with noise. Green dots indicate an in-focus peak returned by the function, while red dots indicate a rejected peak (the function returns `None`). 
Screenshot 2023-07-11 at 11 03 31 AM

**Performance on synthetic data**
Caption: an SNR sweep on simulated data. On average, the peak width for an in-focus object decreases as the SNR decreases, so thresholding on the peak width is an appropriate strategy. 
Screenshot 2023-07-11 at 11 06 29 AM

**Comment on (lack of) danger to hardware**
This approach uses data that's already been acquired to estimate in-focus positions, so it does not introduce any additional danger of crashing. Increasing the `threshold_FWHM` increases the safety since it reduces the likelihood of choosing slices that are at the edge of the axial FOV, which reduces the chance of repeated run leading to a crash. 

**Next steps**
@ieivanov if you find that `threshold_FWHM` does not provide the discriminatory power you need, we can iterate and try different threshold/peak finders.

### DIFF
--- a/tests/test_focus_estimator.py
+++ b/tests/test_focus_estimator.py
@@ -3,8 +3,7 @@ import numpy as np
 from waveorder import focus
 
 
-def test_focus_estimator():
-
+def test_focus_estimator(tmp_path):
     ps = 6.5 / 100
     lambda_ill = 0.532
     NA_det = 1.4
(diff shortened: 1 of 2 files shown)


ieivanov:
Thanks @talonchandler! I'll give this a try and will report back.

As a possible next iteration - I think the MATLAB [findpeaks](https://www.mathworks.com/help/signal/ref/findpeaks.html) does a very good job. [Peak prominence](https://www.mathworks.com/help/signal/ug/prominence.html) is a metric which relates the peak FWHM and the peak height above the "background".

This seems to be a scipy implementation: https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.find_peaks.html

ieivanov:
You're correct. 

I've been thinking some more about this, and it seems to me that the peak prominences may not be the best metric as it would prioritize tall and skinny peaks. In our application we want features which come in and out of focus slowly, as you said, and thus have a wide FWHM. I think something like the peak "area" - width * abs_height, would be a better metric that will keep us away from close calls like in the bottom right. I'll test today and will report back. 

I think the strategy of finding the highest peak and evaluating peak metrics, as we do now, is right, rather than using `peak_find`.

ieivanov:
I've convinced myself that the FWHM is the right parameter here. Let's merge this! Thanks again.

ieivanov (approved):

PR #134, opened 2023-07-12
Remove `__init__` imports

### LINKED ISSUE #133: waveorder changes matplotlib backend at import
The waveorder `__init__.py` file imports all library modules, even if they are not used. This includes `from .visual import *`, which changes the default `matplotlib` backend, in my case to `QtAgg`, even if function from the `visual` module are not used, which can create problems with other libraries. Can we remove the `from .visual import *` line, or even the entire contents of this file? I think this is not a good practice in general.

talonchandler:
Fixes #133 and updates all of the CI-tested examples to use the individual modules instead of a global `wo` namespace. For example, the previous `wo.plot_multicolumn` becomes `visual.plot_multicolumn`. 

Also, removes unused imports from across `waveorder`.

### DIFF
--- a/examples/2D_QLIPP_simulation/2D_QLIPP_forward.py
+++ b/examples/2D_QLIPP_simulation/2D_QLIPP_forward.py
@@ -12,11 +12,10 @@
 
 import numpy as np
 import matplotlib.pyplot as plt
-from numpy.fft import fft2, ifft2, fftshift, ifftshift
+from numpy.fft import fftshift
 from waveorder import (
     optics,
     waveorder_simulator,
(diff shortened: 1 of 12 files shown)


ieivanov (approved):
Looks good to me, thanks @talonchandler. I've only looked over the code, haven't run any of the examples

ieivanov:
Should we merge this into `alg_dev` as well?

talonchandler:
I've run the examples locally and via CI, so I'm ready to merge. I'll merge this into `main` then merge `main` into `alg-dev`.

PR #136, opened 2023-07-18
close figure after saving

ieivanov:
`matplotlib.pyplot` figures remain open until explicitly closed. Calling `focus_from_transverse_band` multiple times keeps open figures in memory - we should close them after saving


talonchandler (approved):
Thanks @ieivanov

PR #141, opened 2023-07-25
Fix readme figure

### LINKED ISSUE #106: README figure is low-res and transparent
The latest README figure is lower resolution than the previous one, so the smallest text is barely legible:

Screen Shot 2023-02-22 at 11 00 20 AM

@ziw-liu also pointed out that the background is mostly transparent and inconsistent:

> the equation in the figure is on a solid white background, while the rest of the figure is on a transparent background that may have contrast issues in a dark theme

@mattersoflight can you point me to the source file or fix these when you get a chance?

Note: this does not need to be fixed before the release, since I'm linking the github file directly (I learned through recOrder releases that linking the github file is a good strategy since distributing .pngs is suboptimal---no need to pip-install big pngs for everyone).

talonchandler:
Fixes #106.

Here I've added a single 400 dpi .png with a white background. 

I tried a raw .svg, but the file was larger, and I wasn't happy with the scaling behavior. 

This .png has a white background, so it is legible in all themes. 

I have a soft preference for a single fixed white background figure instead of maintaining two versions of the same figure (I think converting this to a  black background will require more care than a global filter...some lines should be inverted, but some shouldn't), but I'm open to your preferences/suggestions.


ziw-liu (approved):
Looks good!

PR #148, opened 2023-08-10
Add rotate and flip orientation options to `waveorder` options

talonchandler:
In [`recOrder` #248](https://github.com/mehta-lab/recOrder/pull/248) we added an orientation offset checkbox to `recOrder`'s GUI. 

This PR is a first step towards moving this logic to `waveorder` so that these options are more easily available. 

I've also added a `flip` option so that all combinations are covered. Following this merge, I will remove the business logic from `recOrder`, add another button to the `recOrder` GUI, then connect the `rotation` and `flip` buttons to the waveorder side.

### DIFF
--- a/tests/models/test_inplane_oriented_thick_pol3D.py
+++ b/tests/models/test_inplane_oriented_thick_pol3D.py
@@ -18,7 +18,7 @@ def test_calculate_transfer_function():
 
 def test_apply_inverse_transfer_function():
     input_shape = (5, 10, 5, 5)
-    czyx_data = torch.randn(input_shape)
+    czyx_data = torch.rand(input_shape)
 
     intensity_to_stokes_matrix = (
         inplane_oriented_thick_pol3d.calculate_transfer_function(
--- a/tests/models/test_inplane_oriented_thick_pol3D.py
+++ b/tests/models/test_inplane_oriented_thick_pol3D.py
@@ -1,5 +1,6 @@
 import pytest
 import torch
+
 from waveorder import stokes
 from waveorder.models import inplane_oriented_thick_pol3d
 
@@ -13,3 +14,25 @@ def test_calculate_transfer_function():
     )
 
     assert intensity_to_stokes_matrix.shape == (4, 5)
+
+
+def test_apply_inverse_transfer_function():
+    input_shape = (5, 10, 5, 5)
+    czyx_data = torch.randn(input_shape)
+
+    intensity_to_stokes_matrix = (
+        inplane_oriented_thick_pol3d.calculate_transfer_function(
+            swing=0.1,
+            scheme="5-State",
+        )
+    )
+
+    results = inplane_oriented_thick_pol3d.apply_inverse_transfer_function(
+        czyx_data=czyx_data,
+        intensity_to_stokes_matrix=intensity_to_stokes_matrix,
+    )
+
+    assert len(results) == 4
+
+    for result in results:
+        assert result.shape == input_shape[1:]
--- a/waveorder/stokes.py
+++ b/waveorder/stokes.py
@@ -426,7 +426,7 @@ def mmul(matrix, vector):
 
 def apply_orientation_offset(orientation, rotate, flip):
     """
-    Applies an in-place rotation and/or flip to an orientation map while
+    Applies a rotation and/or flip to an orientation map while
     keeping the output range within 0 <= orientation < pi.
 
     Parameters
--- a/waveorder/stokes.py
+++ b/waveorder/stokes.py
@@ -426,7 +426,7 @@ def mmul(matrix, vector):
 
 def apply_orientation_offset(orientation, rotate, flip):
     """
-    Applies a rotation and/or flip to an orientation map while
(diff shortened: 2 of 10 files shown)


waveorder/stokes.py
```diff
@@ -422,3 +422,37 @@ def mmul(matrix, vector):
...
+
+
+def apply_orientation_offset(orientation, rotate, flip):
+    """
+    Applies an in-place rotation and/or flip to an orientation map while
+    keeping the output range within 0 <= orientation < pi.
+
+    Parameters
+    ----------
+    orientation : array_like
```
ziw-liu, line 434: Does this actually work for array-like, e.g. numpy arrays?

waveorder/stokes.py
```diff
@@ -422,3 +422,37 @@ def mmul(matrix, vector):
...
+
+    Transformed array of orientations measured in radians
+    with range 0 <= orientation < pi
+
+    Note
+    ----
+    rotate=False and flip=False leaves the effective orientation unchanged
+    while changing the output range to 0 <= orientation < pi
+    """
+    out_orientation = torch.clone(orientation)
```
ziw-liu, line 453: Is this copy necessary? The docstring says that this operation is in-place.

talonchandler: Thanks for the comment. I changed the docstring to
"Applies a rotation and/or flip to each voxel of an orientation map..."

My initial intent for "in-place" was to imply that the rotation is voxel-wise, but your comment made me realize that it makes it seem like I'm modifying the whole array in place. 

FWIW I found that I did need to clone if I want to use the (IMO) clean `+=` and `*=` operators. I could skip these operators, but this would end up duplicating anyway. 

I also don't want to modify the input array, because in my experience this causes tough-to-trace downstream bugs.

waveorder/stokes.py
talonchandler: Thanks, I changed `array_like` to `torch.tensor` everywhere.

tests/models/test_inplane_oriented_thick_pol3D.py
```diff
@@ -13,3 +14,25 @@ def test_calculate_transfer_function():
     )
 
     assert intensity_to_stokes_matrix.shape == (4, 5)
+
+
+def test_apply_inverse_transfer_function():
+    input_shape = (5, 10, 5, 5)
+    czyx_data = torch.randn(input_shape)
```
ziw-liu, line 21: Is `randn` needed? A uniform distribution with `rand` should also work.

waveorder/stokes.py
```diff
@@ -147,16 +147,16 @@ def stokes_after_adr(
     depolarizing retarder (adr) parametrized by its retardance, slow-axis
     orientation, transmittance, and depolarization.
 
-    Note: all four parameters can be array_like, but they must be the same size.
-    If your parameters are in an ndarray with array.shape = (4, ...), use
+    Note: all four parameters can be torch.tensor, but they must be the same size.
```
ziw-liu, line 150: Note that `torch.tensor` is a functional constructor of the `torch.Tensor` type. This is different from `numpy.array` which is effectively an alias.

edyoshikun:
I wasn't getting the results I was expecting by adding the `inverse_phase_contrast` parameter to the `examples/model/phase_thick_3d.py`

```python 
transfer_function_arguments = {
    "z_padding": 0,
    "numerical_aperture_illumination": 0.9,
    "numerical_aperture_detection": 1.2,
    "invert_phase_contrast": True,
}
```
When I look at the reconstructions and the steps in between , I was expecting that the transfer function was flipped and that the output had the inverse contrast as well. Here, `reconstruction_2` refers to the`invert_phase_contrast=false`. I want to make sure I was testing this appropriately. 

Screenshot 2023-08-14 at 10 12 20 AM

edyoshikun (dismissed):
The implementation of the orientation flip and phase contrast inversion makes sense to live in waveorder. I also appreciated that these modules default to a bool state and that these can be tweaked with just the waveorder library. 

Also, thank you for checking my previous issue with the simulation. It is now clear to me that given the order of steps we take in the example to generate the simulated dataset and the reconstruction the example is behaving as it is supposed to.

waveorder/stokes.py
talonchandler: Thanks...changing `torch.tensor` -> `torch.Tensor` in the docstrings.

ziw-liu (approved):
Code looks good! I'll leave the data validation to @edyoshikun.

talonchandler:
Thanks @edyoshikun and @ziw-liu. 

For documentation purposes: @edyoshikun's question above was related to a misunderstanding about how the `invert_phase_contrast` parameter works. If you use `invert_phase_contrast` when calculating the transfer function, that transfer function will **simulate** and **reconstruct** as if the data were collected with the opposite z stage order. If you do both a simulation and a reconstruction, then the inversion will cancel and you'll get the same reconstruction with or without `invert_phase_contrast`. 

We decided to keep the current design because the typical user will only be using the reconstruction, and expert users who need both simulations and reconstructions will not use the `invert_phase_contrast`.

PR #159, opened 2024-03-01
Update a single PTI example notebook

talonchandler:
This PR updates a single example notebook `PTI_Experiment_Recon3D_anisotropic_target_small.ipynb` so that it runs with the latest version of the waveorder. It requires a 500 MB download and a ~15-minute single-CPU reconstruction, so I have not added it to the test suite. Additionally, I have documented that this notebook now runs, while the others serve as dcoumentation. 

This updated notebook gives me high confidence that the results in the remaining notebooks are reproducible in principle, though exact reproduction will more extensive migration + computation. 

Together with the [reconstruction example scripts that are part of our automated test suit here](https://github.com/mehta-lab/waveorder/tree/main/examples/maintenance/PTI_simulation), this example notebook serves as a very good jumping-off point for PTI reconstructions.

### DIFF
--- a/examples/documentation/PTI_experiment/README.md
+++ b/examples/documentation/PTI_experiment/README.md
@@ -1,2 +1,6 @@
 The notebooks in this folder require data from this source:
 https://www.ebi.ac.uk/biostudies/bioimages/studies/S-BIAD1063
+
+One notebook `PTI_Experiment_Recon3D_anisotropic_target_small.ipynb` will run as is with the latest version of the code and a ~500 MB download from https://www.ebi.ac.uk/biostudies/files/S-BIAD1063/PTI-BIA/Anisotropic_target_small.zip.
+
+The remaining notebooks are not maintained, so they will require a download and an old version of this repository from https://github.com/mehta-lab/waveorder/tree/1.0.0b0.


ziw-liu:
In 'Load sample images', the `zarr.open` call should use `mode="r"` to prevent creating/writing and provide `FileNotFound` errors if data was not downloaded properly.

ziw-liu:
Low priority: we should run black on these notebooks at some point.

ziw-liu:
> In 'Load sample images', the `zarr.open` call should use `mode="r"` to prevent creating/writing and provide `FileNotFound` errors if data was not downloaded properly.
> 
> edit: different from `iohub.open_ome_zarr()` which defaults to `"r"`, `zarr.open()` defaults to `"a"`

Also here:

```py
# Load the processed results
PTI_file_name = '/path/to/Anisotropic_target_small/Anisotropic_target_small_processed.zarr'
reader = zarr.open(PTI_file_name)
PTI_array = np.array(reader["Row_0/Col_0/f_tensor/array"])
```

ziw-liu (approved):
The notebook runs and outputs visuals. I didn't check the results for correctness though.

ziw-liu (approved):

PR #160, opened 2024-03-05
PTI example jupyter notebook to script and .html

talonchandler:
This PR moves the PTI spoke-pattern data example from a jupyter notebook to a script and a static .html file. 

This PR also includes several suggestions from @mattersoflight to match the names to the most recent PTI submission:  
- [x] Change section title 3D uPTI reconstruction -> 3D PTI reconstruction.
- [x] Change subsection title Reconstructed phase, absorption, principal retardance, azimuth, and inclination assuming (+) and (-) optic sign -> Reconstructed mean permittivity, differential permittivity, azimuth, and inclination assuming (+) and (-) optic sign.  The text should clarify  that the real part of mean permittivity causes phase delay and imaginary part of mean permittivity causes absorption.
- [x] Figure title for phase (xz) reads absorption (xz).
- [x] Figure title principal retardance -> differential permittivity
- [x] @ziw-liu I've applied `black` to the notebook-style script

Some other TODOs/questions:
- [x] Decide if we want a .pdf instead of .html. My first attempt at a .pdf led to TexLive issues together with font- and figure-size weirdness. The .html worked well out of the box. I'm open to suggestions and constraints I'm not thinking of.  
- [x] Delete the jupyter notebook before merging. I've kept it temporarily since it's the source of the other two files and I may need to change them all together in this PR.

### DIFF
--- a/examples/documentation/PTI_experiment/PTI_Experiment_Recon3D_anisotropic_target_small.py
+++ b/examples/documentation/PTI_experiment/PTI_Experiment_Recon3D_anisotropic_target_small.py
@@ -39,7 +39,7 @@ gpu_id = 0  # id of gpu to use
 # %%
 # Load data and bg
 # Download data from
-PTI_file_name = "/Users/talon.chandler/Downloads/Anisotropic_target_small/Anisotropic_target_small_raw.zarr"
+PTI_file_name = "/Users/shalin.mehta/docs/data/waveOrder/Anisotropic_target_small/Anisotropic_target_small_raw.zarr"
 reader = zarr.open(PTI_file_name, mode="r")
 I_meas = np.transpose(
     np.array(reader["Row_0/Col_0/I_meas/array"]), (0, 1, 3, 4, 2)
@@ -305,24 +305,6 @@ visual.parallel_4D_viewer(
     ],
 )
 
-# %%
-# (Obsolete, not maintained) save results to zarr array
-
-# writer = WaveorderWriter('.', hcs=False, hcs_meta=None, verbose=True)
-# writer.create_zarr_root('Anisotropic_target_small_processed.zarr')
-# chan_names = ['f_tensor0r', 'f_tensor0i', 'f_tensor1c','f_tensor1s','f_tensor2c','f_tensor2s', 'f_tensor3', 'mat_map0', 'mat_map1']
-# PTI_array = np.transpose(np.concatenate((f_tensor, mat_map),axis=0)[np.newaxis,...],(0,1,4,2,3)) # dimension (T, C, Z, Y, X)
-# data_shape = PTI_array.shape
-# chunk_size = (1,1,1)+PTI_array.shape[3:]
-# writer.init_array(0, data_shape, chunk_size, chan_names, position_name='f_tensor', overwrite=True)
-# writer.write(PTI_array, p=0)
-
-# chan_names_phys = ['Phase3D', 'Retardance3D', 'Orientation', 'Inclination', 'Optic_sign']
-# phys_data_array = np.transpose(np.array([phase_PT, np.abs(retardance_pr_PT[1]), azimuth[1], theta[1], p_mat_map]),(0,3,1,2))[np.newaxis,...]
-# data_shape_phys = phys_data_array.shape
-# dtype = 'float32'
-# writer.init_array(1, data_shape_phys, chunk_size, chan_names_phys, dtype, position_name='Stitched_physical', overwrite=True)
-# writer.write(phys_data_array, p=1)
 
 # %%
 # Load the processed results
--- a/examples/documentation/PTI_experiment/PTI_Experiment_Recon3D_anisotropic_target_small.py
+++ b/examples/documentation/PTI_experiment/PTI_Experiment_Recon3D_anisotropic_target_small.py
@@ -1,0 +1,695 @@
+# %%
+# This notebook-style script requires a ~500 MB download from https://www.ebi.ac.uk/biostudies/files/S-BIAD1063/PTI-BIA/Anisotropic_target_small.zip
+
+import numpy as np
+import matplotlib.pyplot as plt
+from numpy.fft import fftshift
+
+import waveorder as wo
+from waveorder import optics, waveorder_reconstructor, util, visual
+
+import zarr
+
+# %%
+# Initialization
+
+# Experimental parameters
+# %%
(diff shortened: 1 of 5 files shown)


talonchandler:
One disadvantage of the .html file is that github doesn't render it, and there doesn't seem to be an easy way to open it in another tab. For comparison here's a saved .pdf of the .html file.  
[PTI_Experiment_Recon3D_anisotropic_target_small.pdf](https://github.com/mehta-lab/waveorder/files/14501688/PTI_Experiment_Recon3D_anisotropic_target_small.pdf)

ziw-liu:
GitHub doesn't render it because it's too large. But we can link to a third party one e.g. 
https://html-preview.github.io/?url=https://github.com/mehta-lab/waveorder/blob/a93fd512c6d66cb1b4bd8b983b288195cbb67fe5/examples/documentation/PTI_experiment/PTI_Experiment_Recon3D_anisotropic_target_small.html

mattersoflight:
@talonchandler @ziw-liu thanks! Are you using vscode's Jupyter extension for PDF export? Does it require TexLive in the backend?
image

I get the following PDF after running two cells.
[test.pdf](https://github.com/mehta-lab/waveorder/files/14528588/test.pdf)

I will be able to do the proper review of the script tomorrow and can export a more complete PDF in the process.  @ziw-liu's solution of rendering via https://html-preview.github.io/ also works. We can link the HTML in the README via that service.

talonchandler:
Aha, thanks. I got `jupyter nbconvert` (what vscode's convert uses) to run ([I needed `conda install pandoc`](https://github.com/ipython/ipython/issues/12238)) and the pdf looks good. I think I prefer it to the html, so I updated to the pdf.

examples/documentation/PTI_experiment/PTI_Experiment_Recon3D_anisotropic_target_small.py
```diff
@@ -0,0 +1,695 @@
...
+PTI_file_name = (
+    "/path/to/Anisotropic_target_small/Anisotropic_target_small_processed.zarr"
+)
+reader = zarr.open(PTI_file_name, mode="r")
+PTI_array = np.array(reader["Row_0/Col_0/f_tensor/array"])
+print(PTI_array.shape)
+PTI_array = np.transpose(PTI_array, (0, 1, 3, 4, 2))[0]
+f_tensor = PTI_array[:7]
+mat_map = PTI_array[7:]
+
```
mattersoflight, line 338: Can we assign names to these channels? I understand the previous version of iohub also had an API for that.

examples/documentation/PTI_experiment/PTI_Experiment_Recon3D_anisotropic_target_small.py
```diff
@@ -0,0 +1,677 @@
...
+# Load the processed results
+PTI_file_name = (
+    "/path/to/Anisotropic_target_small/Anisotropic_target_small_processed.zarr"
+)
+reader = zarr.open(PTI_file_name, mode="r")
+PTI_array = np.array(reader["Row_0/Col_0/f_tensor/array"])
+print(PTI_array.shape)
+PTI_array = np.transpose(PTI_array, (0, 1, 3, 4, 2))[0]
+f_tensor = PTI_array[:7]
+mat_map = PTI_array[7:]
```
mattersoflight, line 319: @talonchandler can you please assign names to the channels, so that we can visualize the processed data by loading in napari? I think the released version of iohub has an API for that.

examples/documentation/PTI_experiment/PTI_Experiment_Recon3D_anisotropic_target_small.py
```diff
@@ -0,0 +1,677 @@
...
+        r"$\omega$ (+)",
+        r"$\theta$ (+)",
+        r"absorption",
+        r"mean permittivity (-)",
+        r"$\omega$ (-)",
+        r"$\theta$ (-)",
+    ],
+)
+
+
```
mattersoflight, line 318: @talonchandler this is just a stub - attempt to replace the waveorderWriter w/ iohub. Can you complete and test? 
I suggest we save scattering potential tensor and reconstructed physical properties assuming the (-) optic sign to separate zarr store. There are many "channels" here. It will be very useful to map the channel names to the quantities in the paper for our future selves.

examples/documentation/PTI_experiment/PTI_Experiment_Recon3D_anisotropic_target_small.py
```diff
@@ -0,0 +1,677 @@
...
+            ((-1) ** i) * mean_permittivity_PT[i], "db8", 0.00303, level=1
+        )
+        for i in range(2)
+    ]
+)
+
+
+# %%
+# Phase, mean permittivity, azimuth, inclination, and optic sign reconstruction
+
```
mattersoflight, line 373: @talonchandler looks like you read that mean permittivity is equivalent to principal retardance. It is actually the differential permittivity. I was partial to isotropic permittivity and anisotropic permittivity, but this is the terminology we have now throughout the figures and videos. Can you check the variable names in previous section and this section?

mattersoflight (commented):
Thanks for this iteration @talonchandler.  The combination of the script and PDF works well for demo and documentation. Feel free to delete the old notebook that we don't want to maintain. The names of the variables in the demo script need to be corrected and we should switch to using iohub to wite the zarr store.

mattersoflight (changes requested):
Thanks for this iteration @talonchandler.  The combination of the script and PDF works well for demo and documentation. Feel free to delete the old notebook that we don't want to maintain. The names of the variables in the demo script need to be corrected and we should switch to using iohub to wite the zarr store.

talonchandler:
Thanks Shalin:

- [x] reconstructions should save to napari-readable zarr with channel names (see attached screenshot)
- [x] principal retardance should be differential permittivity (thanks, pardon my mistake)
- [x] don't track the jupyter notebook

 
Screenshot 2024-03-11 at 5 57 27 PM

talonchandler:
P.S. I tested that all of the cells run...only thing that needs to change is the path to the data.

mattersoflight (approved):
Thanks @talonchandler all set to merge and release 2.1.0! 
I tested that the OME-zarr store opens in napari.

talonchandler:
Thanks for the fixes! Merging and releasing now.

PR #162, opened 2024-03-18
fix bug finding focus in stack with only one slice

ieivanov:
Fixing bug related to focus finding in stack with only one slice. `in_focus_index=0` should be returned by `focus_from_transverse_band`, not `_check_focus_inputs`. `_check_focus_inputs` could also be refactored into two functions that 1. check the inputs and 2. determine the correct optimization function. This will remove possible confusion on `_check_focus_inputs` returning a callable.

### DIFF
--- a/tests/test_focus_estimator.py
+++ b/tests/test_focus_estimator.py
@@ -45,6 +45,15 @@ def test_focus_estimator(tmp_path):
     assert slice <= data3D.shape[0]
     assert plot_path.exists()
 
+    # Check single slice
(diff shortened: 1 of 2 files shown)


talonchandler:
Thanks for catching and fixing, @ieivanov. I followed your refactoring suggestion, and I moved the single-slice check to the top level so that it can return 0 directly.

talonchandler (approved):

ieivanov:
Looks good! Merging.

PR #176, opened 2024-10-21
Bump torch to unpin numpy

ziw-liu:
https://github.com/pytorch/pytorch/issues/128860 has been fixed in torch 2.4.1. The temporary pin in #173 is no longer necessary.

Edit: also bump the numpy requirement as recommended in [SPEC-0](https://scientific-python.org/specs/spec-0000/#2024---quarter-4).


talonchandler (approved):
Thanks @ziw-liu

PR #178, opened 2024-11-08
Transfer function visuals

talonchandler:
This PR adds visualization scripts for the Green's tensor spectrum and transfer functions. 

Tagging @mattersoflight, @edyoshikun, @ieivanov who have requested these.  

Running these scripts results in:

https://github.com/user-attachments/assets/184b48c2-764b-4ea7-b649-05da67a96035

https://github.com/user-attachments/assets/8f3ac2d3-ece1-4913-95ad-31c9562abd13

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,12 +45,12 @@ classifiers = [
   "Operating System :: MacOS",
 ]
 dependencies = [
-  "numpy>=1.21, <2",
+  "numpy>=1.24",
   "matplotlib>=3.1.1",
   "scipy>=1.3.0",
   "pywavelets>=1.1.1",
   "ipywidgets>=7.5.1",
-  "torch>=2.2.1",
+  "torch>=2.4.1",
 ]
 dynamic = ["version"]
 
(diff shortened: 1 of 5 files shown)


mattersoflight (approved):
Hi @talonchandler, these scripts work well! I tested on Mac. 

Surface rendering of the support of the Green's tensor spectrum just like you surface render support of OTFs will lead to a more intuitive and compelling visualization.

Minor point: For visualization and slides, the dark theme looks better to me.

image

talonchandler:
>Surface rendering of the support of the Green's tensor spectrum just like you surface render support of OTFs will lead to a more intuitive and compelling visualization.

I agree, I will iterate here. The aliasing and colormapping are making the volume renders suboptimal. I'll need to get a "marching cubes" or similar algorithm connected to napari's surface layer, but I don't expect an enormous lift.  

> Minor point: For visualization and slides, the dark theme looks better to me.
Sounds good, I can return to that default after changing to a surface rendered Green's.

talonchandler:
- Dark mode is now default for these scripts

- Green's tensor is now rendered as a surface (see screenshots below)
Screenshot 2024-12-05 at 5 17 08 PM


- The paper figure is now updated with the surface renders
Screenshot 2024-12-05 at 5 36 10 PM

PR #180, opened 2025-01-31
Calculate transfer functions on grids smaller than data  

talonchandler:
This PR allows transfer functions to be calculated on smaller grids then applied to larger grids.  

This PR uses a `stretched_multiply` strategy for efficient application of a small OTF to a large dataset. This approach approximates the OTF as a piece-wise constant function, and applies this smaller array to datasets without upsampling. This approach allows us to reduce OTF memory requirements by as much as needed, trading OTF computational expense and OTF memory for accuracy.

To uniformly apply this strategy, I have implemented a new function `filter.apply_filter_bank`, which handles these the prepadding, stretched multiplications, matrix-vector bank multiplications, and unpadding. 

These three reconstructions now use a precomputed inverse filter bank and use the new shared `filter.apply_filter_bank`. 
- [x] isotropic_thick_3d
- [x] isotropic_fluorescent_thick_3d
- [x] inplane_oriented_thick_pol3d_vector

### DIFF
--- a/tests/test_filter.py
+++ b/tests/test_filter.py
@@ -36,10 +36,8 @@ def test_stretched_multiply():
 
     # Test that output dims are correct
     rand_array_3x3x3 = torch.rand((3, 3, 3))
-    rand_array_100x100x100 = torch.rand((99, 99, 99))
-    result = filter.stretched_multiply(
-        rand_array_3x3x3, rand_array_100x100x100
-    )
+    rand_array_99x99x99 = torch.rand((99, 99, 99))
+    result = filter.stretched_multiply(rand_array_3x3x3, rand_array_99x99x99)
     assert result.shape == (99, 99, 99)
 
 
--- a/tests/test_filter.py
+++ b/tests/test_filter.py
@@ -4,17 +4,19 @@ import pytest
 
 
 def test_apply_transfer_function_filter():
-    input_array = torch.tensor([[1.0, 2.0], [3.0, 4.0]])
-    transfer_function = torch.tensor([[1, 0], [0, 0]])
-    result = filter.apply_transfer_function_filter(transfer_function, input_array)
-    expected = torch.tensor([[10, 10], [10, 10]])/4
+    input_array = torch.tensor([[[1.0, 2.0], [3.0, 4.0]]])
+    transfer_function_bank = torch.tensor([[[[1, 0], [0, 0]]]])
+    result = filter.apply_filter_bank(transfer_function_bank, input_array)
+    expected = torch.tensor([[[10, 10], [10, 10]]]) / 4
     assert torch.allclose(result, expected)
 
     # Test with incompatible shapes
-    input_array = torch.tensor([[1.0, 2.0], [3.0, 4.0]])
-    transfer_function = torch.tensor([[0.5, 0.5, 0.5], [0.5, 0.5, 0.5]])
+    input_array = torch.tensor([[[1.0, 2.0], [3.0, 4.0]]])
+    transfer_function_bank = torch.tensor(
+        [[[[0.5, 0.5, 0.5], [0.5, 0.5, 0.5]]]]
+    )
     with pytest.raises(ValueError):
-        filter.apply_transfer_function_filter(transfer_function, input_array)
+        filter.apply_filter_bank(transfer_function_bank, input_array)
 
 
 def test_stretched_multiply():
--- a/tests/test_filter.py
+++ b/tests/test_filter.py
@@ -1,0 +1,42 @@
+from waveorder import filter
+import torch
+import pytest
+
+
+def test_stretched_multiply():
+    small_array = torch.tensor([[1, 2], [3, 4]])
+    large_array = torch.tensor(
+        [[1, 2, 3, 4], [5, 6, 7, 8], [9, 10, 11, 12], [13, 14, 15, 16]]
+    )
+    result = filter.stretched_multiply(small_array, large_array)
+    expected = torch.tensor(
+        [[1, 2, 6, 8], [5, 6, 14, 16], [27, 30, 44, 48], [39, 42, 60, 64]]
+    )
+    assert torch.all(result == expected)
+    assert torch.all(
+        filter.stretched_multiply(large_array, large_array) == large_array**2
+    )
+
+    # Test that output dims are correct
+    rand_array_3x3x3 = torch.rand((3, 3, 3))
+    rand_array_100x100x100 = torch.rand((100, 100, 100))
+    result = filter.stretched_multiply(
+        rand_array_3x3x3, rand_array_100x100x100
+    )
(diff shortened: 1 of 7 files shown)


waveorder/filter.py
```diff
@@ -0,0 +1,196 @@
+import itertools
+import numpy as np
```
ziw-liu, line 2: unused import

ziw-liu: Sidenote: we should add linting for this.

waveorder/filter.py
```diff
@@ -0,0 +1,196 @@
...
+            "io_filter_bank and i_input_array must have the same number of spatial dimensions."
+        )
+
+    # Ensure the input dimensions match
+    if io_filter_bank.shape[0] != i_input_array.shape[0]:
+        raise ValueError(
+            "io_filter_bank.shape[0] and i_input_array.shape[0] must be the same."
+        )
+
+    I, O = io_filter_bank.shape[:2]
```
ziw-liu, line 74: What is `I` and `O`? Is it possible to use lower case readable words for these variables?

waveorder/filter.py
```diff
@@ -0,0 +1,196 @@
...
+    padded_output_spectrum = torch.zeros((O,) + spatial_dims, dtype=padded_input_spectrum.dtype)
+    for i_idx in range(I):
+        for o_idx in range(O):
+            padded_output_spectrum[o_idx] += stretched_multiply(
+                io_filter_bank[i_idx, o_idx], padded_input_spectrum[i_idx]
+            )
+
+    # Casts to input_array dtype, which typically ignores imaginary part
+    padded_result = torch.fft.ifftn(padded_output_spectrum, dim=fft_dims).type(
+        i_input_array.dtype
```
ziw-liu, line 104: What is the expected type of `i_input_array`?

ziw-liu:
> These three reconstructions now use a precomputed inverse filter bank

Should this also apply to `isotropic_thin_3d`?

waveorder/filter.py
talonchandler: Fixed in #182.

waveorder/filter.py
talonchandler: changed `I` -> `num_input_channels` and `O` -> `num_output_channels`. Thanks!

waveorder/filter.py
talonchandler: Thanks for flagging. 

I was being more general than needed here...we only need real-valued input and outputs for `i_input_array`, and I've updated the docs to make this clearer.

talonchandler:
ZL> Should this also apply to `isotropic_thin_3d`?

Good question. I have added a TODO for this, but I am going to put it off for now because `isotropic_thin_3d` is undergoing a separate debugging effort and I want to avoid muddying the waters.

tests/test_filter.py
```diff
@@ -0,0 +1,59 @@
...
+        [[1, 2, 6, 8], [5, 6, 14, 16], [27, 30, 44, 48], [39, 42, 60, 64]]
+    )
+    assert torch.all(result == expected)
+    assert torch.all(
+        filter.stretched_multiply(large_array, large_array) == large_array**2
+    )
+
+    # Test that output dims are correct
+    rand_array_3x3x3 = torch.rand((3, 3, 3))
+    rand_array_100x100x100 = torch.rand((99, 99, 99))
```
ziw-liu, line 39: Why is this named 100 when the size is 99?

waveorder/filter.py
```diff
@@ -0,0 +1,202 @@
...
+
+    # Apply the transfer function in the frequency domain
+    fft_dims = [d for d in range(1, i_input_array.ndim)]
+    padded_input_spectrum = torch.fft.fftn(padded_input_array, dim=fft_dims)
+
+    # Matrix-vector multiplication over f
+    # If this is a bottleneck, consider extending `stretched_multiply` to
+    # a `stretched_matrix_multiply` that uses an call like
+    # torch.einsum('io..., i... -> o...', io_filter_bank, padded_input_spectrum)
+    padded_output_spectrum = torch.zeros(
```
ziw-liu, line 97: Does this need a `device=padded_input_spectrum.device` to work on the GPU?

waveorder/filter.py
```diff
@@ -0,0 +1,202 @@
...
+        (num_output_channels,) + spatial_dims,
+        dtype=padded_input_spectrum.dtype,
+    )
+    for input_channel_idx in range(num_input_channels):
+        for output_channel_idx in range(num_output_channels):
+            padded_output_spectrum[output_channel_idx] += stretched_multiply(
```
ziw-liu, line 103: Torch has some special accelerated calls for block-wise multiply-add. But maybe speed is not a bottleneck here.

tests/test_filter.py
talonchandler: Good catch! Vestige from an earlier design. Fixed.

waveorder/filter.py
talonchandler: Good point. Haven't tested end-to-end yet, but can't hurt. Fixed.

waveorder/filter.py
talonchandler: In my latest tests this is currently not a bottleneck (we make at most 3 input x 4 output = 12 trips through this loop), but it could become a bottleneck in the future. 

Still I'd appreciate a link to the block-wise multiply-adds that you have in mind. I'm seeing this [`bmm`](https://pytorch.org/docs/stable/generated/torch.bmm.html), but that's not quite what I'd need here.

ziw-liu: I was thinking of something like [baddmm](https://pytorch.org/docs/stable/generated/torch.baddbmm.html#torch.baddbmm) with reshaping. But I know realize that the 'blocks' here are super small (2x2?) so the memory access pattern is not really 'blocky'. The current approach is similar to pixel-[shuffle](https://pytorch.org/docs/stable/generated/torch.nn.functional.pixel_shuffle.html#torch.nn.functional.pixel_shuffle)/[unshuffle](https://pytorch.org/docs/stable/generated/torch.nn.functional.pixel_unshuffle.html#torch.nn.functional.pixel_unshuffle) for stretched-multiply followed by a point-wise convolution done here.

ziw-liu: (And point-wise convolution is really just a different way to spell 'nn.Linear')

ziw-liu: Can it be done via pixel-shuffle -> linear -> pixel-unshuffle?

talonchandler: Thanks @ziw-liu. After a quick in-person chat we agreed that this is likely premature optimization, but we agreed that there is likely additional optimization available here.  

I've added a comment that points to these functions so that we can optimize further if we find that this is a bottleneck.

ziw-liu (approved):

PR #182, opened 2025-02-18
Styling 

talonchandler:
Styling inspired by `iohub`. Includes `isort` and `black`, but skips `flake8` (lots of work to get `flake8` to pass here).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,62 @@
+# Contributing guide
+
+Thanks for your interest in contributing to `waveorder`!
+
+Please see the following steps for our workflow.
+
+## Getting started
+
+Please read the [README](./README.md) for an overview of the project,
+and how you can install and use the package.
+
+## Issues
+
+We use [issues](https://github.com/mehta-lab/waveorder/issues) to track
+bug reports, feature requests, and provide user support.
+
+Before opening a new issue, please first search existing issues (including closed ones),
+to see if there is an existing discussion about it.
+
+### Setting up development environment
+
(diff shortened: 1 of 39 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,62 @@
...
+cd waveorder/ # or the renamed project root directory
+pip install -e ".[dev]"
+```
+
+Then make the changes and [track them with Git](https://docs.github.com/en/get-started/using-git/about-git#example-contribute-to-an-existing-repository).
+
+
+### Code style
+
+We use [pre-commit](https://pre-commit.com/) to sort imports with [isort](https://github.com/PyCQA/isort), format code with [black](https://black.readthedocs.io/en/stable/), and lint with [flake8](https://github.com/PyCQA/flake8) automatically prior to each commit. To minimize test errors when submitting pull requests, please install pre-commit in your environment as follows:
```
ziw-liu, line 56: Maybe not mention flake8 if it's not used? Also ruff may be a better choice if starting fresh. We use flake8 in iohub because ruff was not around then.

talonchandler: Thanks @ziw-liu. I removed mention of flake8. 

I just tried `ruff`, and I ran into the same issues as flake8 --- it flags (but can't autofix) a large number of issues with the existing code base (especially PTI code), and I do not think it is worth the time investment to fix these issues. For now I suggest we stick with `black` and `isort`, and we reconsider a linter when the PTI code is more stable.

ziw-liu (approved):

Issue #264, opened 2023-02-14
TriggerScope/DAC installation instructions


talonchandler:
Our current [installation guide](https://github.com/mehta-lab/recOrder/blob/main/docs/microscope-installation-guide.md) doesn't include any details about the TriggerScope/DAC. 

@edyoshikun @ieivanov I haven't installed a TriggerScope from scratch, and from discussions with @edyoshikun I know there are some specific TriggerScope MM settings that need to be correct. Would either of you mind adding a short section?

Fairly low priority: I don't think we have any external users installing `recOrder` with a TriggerScope in the near term.

edyoshikun:
I realized I never pushed the draft I had written for this some time ago... It certainly needs images and clarity in the instructions, but this gives us a headstart.

PR #442, opened 2025-03-08
Markdown for triggerscope control

edyoshikun:
Addresses #264 by adding the triggerscope-LC setup guide

- [ ] missing photos of connectors (?)
- [ ] screenshots of the triggerscope + LC + micromanger voltages


edyoshikun:
image
 something like this?

PR #445, opened 2025-04-25
bug in monitor

ahillsley:
bug in monitor, where do_print is passed as a positional rather than a keyword arg, was causing print_indices to contain the wrong variable

### DIFF
--- a/waveorder/cli/monitor.py
+++ b/waveorder/cli/monitor.py
@@ -117,7 +117,7 @@ def monitor_jobs(
 
     # print all jobs once if terminal is too small
     if shutil.get_terminal_size().lines - NON_JOB_LINES < len(jobs):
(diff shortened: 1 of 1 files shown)


talonchandler:
Thanks @ahillsley!

Additional context: this only seems to be a problem when a large number of jobs are submitted. Both @ahillsley and I have run a smaller number of FOVs successfully without this change. 

E.g. see this stack track from @ahillsley:


I'm cc'ing @amitabhverma, who introduced `do_print`. Do you understand what's happening here @amitabhverma?

ahillsley:
Re-raising this issue. Can we approve the PR and merge?

talonchandler (approved):
Thanks @ahillsley. Merging.

Issue #197, opened 2025-01-03
Dependencies on `iohub`, `pydantic`, and `micromanager` are getting stale


talonchandler:
We've started to run into some `iohub`, `pydantic`, and `micromanager` versioning issues in our environments, so I think it's a good time to upgrade these dependencies. 

- I'm planning to upgrade to `iohub==0.2.0a1` (`main` is being used by `biahub`, so I think it's safe to upgrade). 
- Upgrading to `pydantic>=2` is overdue. I'm planning to use the v1 namespace.
- I'll upgrade to a latest nightly `micromanager`, and I'll test the calibrations.

talonchandler:
When we update our `iohub` dependency, we should also upgrade to `iohub.ngff.utils.create_empty_plate`. See https://github.com/mehta-lab/recOrder/pull/461#issuecomment-2708403926. Thanks for flagging @edyoshikun.

ziw-liu:
Is this being worked on? I'm interested in getting feedback on the iohub 0.2.0 release candidate.

talonchandler:
Tagging @ahillsley, who is waiting on this branch to integrate into an OPS environment.

talonchandler:
@ziw-liu has taken on the `iohub` and `pydantic` upgrades in #448, where the tests now pass. Thanks @ziw-liu. 

I will leave this issue open until the `micromanager` upgrade is addressed, currently planned for @amitabhverma.

PR #446, opened 2025-05-02
demos for AQLM 2025

mattersoflight:
I am adding two demos in the `examples/demos/` folder. These are based on simulations on `examples/maintanence` folder. During AQLM, we will finetune these demos. It would make sense to turn the `examples/maintenance` into non-interactive tests after we merge this PR. 

- [x] Make jupytext dependency of dev installation.
- [x] QLIPP forward simulation and reconstruction
- [x] test QLIPP forward simulation and reconstruction using Google Colab: https://colab.research.google.com/github/mehta-lab/waveorder/blob/demos-aqlm/docs/examples/demos/QLIPP/QLIPP_simulation.ipynb 
- [x] QPI from defocus simulation and reconstruction
- [x] test QPI from defocus simulation and reconstruction using Google Colab: 
https://colab.research.google.com/github/mehta-lab/waveorder/blob/demos-aqlm/docs/examples/demos/QPI_defocus/QPI_defocus_simulation.ipynb
- [x] QPI from defocus reconstruction with experimental data 
- [x] test QPI from defocus reconstruction with experimental data https://colab.research.google.com/github/mehta-lab/waveorder/blob/demos-aqlm/docs/examples/demos/QPI_defocus/QPI_defocus_experiment.ipynb

### DIFF
--- a/docs/examples/demos/QLIPP/QLIPP_simulation.py
+++ b/docs/examples/demos/QLIPP/QLIPP_simulation.py
@@ -1,0 +1,300 @@
+# -*- coding: utf-8 -*-
+# ---
+# jupyter:
+#   jupytext:
+#     formats: py:percent,ipynb
+#     text_representation:
+#       extension: .py
+#       format_name: percent
+#       format_version: '1.3'
+#       jupytext_version: 1.15.2
+#   kernelspec:
+#     display_name: Python 3
+#     language: python
+#     name: python3
+# ---
+
+# %% [markdown]
+"""
+# 2D QLIPP Simulation and Reconstruction Demo
+
+This notebook demonstrates forward simulation and reconstruction for Quantitative Label-free Imaging with Phase and Polarization (QLIPP).
+The simulation and reconstruction are based on the QLIPP paper ([here](https://elifesciences.org/articles/55502)):
+
+S.-M. Guo, L.-H. Yeh, J. Folkesson, I. E. Ivanov, A. P. Krishnan, M. G. Keefe, E. Hashemi,
+D. Shin, B. B. Chhun, N. H. Cho, M. D. Leonetti, M. H. Han, T. J. Nowakowski, S. B. Mehta,
+"Revealing architectural order with quantitative label-free imaging and deep learning,"
+eLife 9:e55502 (2020).
+"""
+
+# %% [markdown]
+"""
+## Setup and Imports
+First, let's install the latest version of waveorder from the main branch
+"""
+
+# %%
+import sys
+import subprocess
+
+# Install latest waveorder from main branch
+subprocess.check_call(
+    [
+        sys.executable,
+        "-m",
+        "pip",
+        "install",
+        "git+https://github.com/mehta-lab/waveorder.git@main",
+    ]
+)
+
+# %%
+from pathlib import Path
+
+import matplotlib.pyplot as plt
+import numpy as np
+from numpy.fft import fftshift
+from platformdirs import user_data_dir
+
+from waveorder import optics, util, waveorder_simulator
+from waveorder.visuals import jupyter_visuals
+
+# %% [markdown]
+"""
+## Forward Simulation
+Here we simulate QLIPP measurements of a Siemens star pattern with uniform phase, uniform retardance, and radial orientation.
+"""
+
+# %%
+# Key parameters
+N = 256  # number of pixel in y dimension
+M = 256  # number of pixel in x dimension
+mag = 40  # magnification
+ps = 6.5 / mag  # effective pixel size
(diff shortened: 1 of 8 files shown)


talonchandler:
I've given these notebooks a pass, and they now run end to end locally. 

The QPI simulation notebook runs on Colab, but I'm still trying to figure out `iohub` installation issues in Colab for the QPI experiment notebook.

talonchandler:
Found a solution in `!pip install iohub==0.2.0`. 

FYI @edyoshikun @ziw-liu, `iohub==0.1.0` doesn't seem to be working with colab.

ziw-liu:
> Found a solution in `!pip install iohub==0.2.0`.
> 
> FYI @edyoshikun @ziw-liu, `iohub==0.1.0` doesn't seem to be working with colab.

I think the kernel needs to be restarted for the old version. But we should really update waveorder's dependencies. (#197)

mattersoflight:
@talonchandler, the demos via google colab worked well. I updated the defocus demo to plot 3D OTF, which sparked good discussion. 

Few observations and questions (I am getting familiar again with codebase and will continue to do so as we wrap up the paper):
*  QLIPP reconstruction is done through `waveorder_simulator.waveorder_microscopy_simulator`, which appears to be the code path from earlier (PTI submission). Should we switch to using `waveorder.models.inplane_oriented_thick_pol3d_vector`?
* Is the code for inverse filter shared between above or different?

mattersoflight:
@talonchandler why do we pin iohub to 0.1.0? 

This cell shows this error - but the notebook runs fine.

```
import sys
import subprocess

# Install latest waveorder from main branch
subprocess.check_call(
    [
        sys.executable,
        "-m",
        "pip",
        "install",
        "git+https://github.com/mehta-lab/waveorder.git@main",
    ]
)
!pip install iohub==0.2.0
``` 

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
waveorder 2.2.2.dev1378+g8315e00 requires iohub==0.1.0, but you have iohub 0.2.0 which is incompatible.
waveorder 2.2.2.dev1378+g8315e00 requires pydantic==1.10.19, but you have pydantic 2.11.4 which is incompatible.
Successfully installed iohub-0.2.0 numcodecs-0.15.1 pydantic-2.11.4 pydantic_extra_types-2.10.4 tifffile-2025.3.30 zarr-2.18.7
```

docs/examples/demos/QLIPP/QLIPP_simulation.py
```diff
@@ -0,0 +1,300 @@
...
+# %%
+# Diffraction aware reconstruction assuming slowly varying transmission
+S1_stack = S_image_recon[1].copy() / S_image_recon[0].mean()
+S2_stack = S_image_recon[2].copy() / S_image_recon[0].mean()
+
+# Tikhonov regularization
+retardance, azimuth = setup.Birefringence_recon_2D(
+    S1_stack, S2_stack, method="Tikhonov", reg_br=1e-2
+)
+
```
mattersoflight, line 266: Add a block on phase reconstruction as well.

mattersoflight: Perhaps using `waveorder.models.inplane_oriented_thick_pol3d_vector`

talonchandler:
> QLIPP reconstruction is done through waveorder_simulator.waveorder_microscopy_simulator, which appears to be the code path from earlier (PTI submission). Should we switch to using `waveorder.models.inplane_oriented_thick_pol3d_vector`?

That's correct. As much as possible we should use the code paths in `/docs/examples/models/`.

> Is the code for inverse filter shared across above demos?

The reconstructions in `/docs/examples/models/` use a shared code path for reconstructions wherever possible---e.g. `waveorder.filter.apply_filter_bank` applies all of our Fourier reconstructions.

> why do we pin iohub to 0.1.0?

No longer needed! Thanks @ziw-liu --- see #448.

---

@mattersoflight let's catch up on our plans here after your deadlines. 

- There are major code paths that can be cut soon if we want to prioritize the PTI path. 
- I'd like to chat about which types of notebooks are critical to maintain, and if we can reduce duplication between the notebooks and the napari-based test scripts in `/docs/examples/models`. FWIW these test scripts are my day-to-day entry point into these models at the python level.

talonchandler:
Adding a couple notes following our discussion @mattersoflight:
- we think notebooks generated by jupytext from scripts are a good way to 
- we've found that installing napari in a conda environment is a significant barrier to entry. We want to try sharing a docker container as a way for a new user to run our scripts with napari.

PR #448, opened 2025-05-13
Bump iohub and pydantic

ziw-liu:
Partial fix for #197 (not including pycro-manager).

### DIFF
--- a/tests/cli_tests/test_cli.py
+++ b/tests/cli_tests/test_cli.py
@@ -7,5 +7,5 @@ def test_main():
     runner = CliRunner()
     result = runner.invoke(cli)
 
-    assert result.exit_code == 0
+    assert result.exit_code == 2
     assert "Toolkit" in result.output
--- a/waveorder/acq/acq_functions.py
+++ b/waveorder/acq/acq_functions.py
@@ -4,7 +4,7 @@ import os
 import time
 
 import numpy as np
-from iohub import read_images
+from iohub.mmstack import MMStack
 
 try:
     from pycromanager import Studio
@@ -127,7 +127,7 @@ def acquire_from_settings(
     settings: dict,
     grab_images: bool = True,
     restore_settings: bool = True,
-):
+) -> np.typing.NDArray:
     """Function to acquire an MDA acquisition with the native MM MDA Engine.
     Assumes single position acquisition.
 
@@ -163,9 +163,4 @@ def acquire_from_settings(
         path = os.path.join(settings["root"], settings["prefix"])
         files = glob.glob(path + "*")
         index = max([int(x.split(path + "_")[1]) for x in files])
-
-        reader = read_images(
-            path + f"_{index}", data_type="ometiff", extract_data=True
-        )
-
-        return reader[0].xdata.to_numpy()
+        return MMStack(path + f"_{index}")[0].xdata.to_numpy()
--- a/waveorder/acq/acq_functions.py
+++ b/waveorder/acq/acq_functions.py
@@ -4,7 +4,7 @@ import os
 import time
 
 import numpy as np
-from iohub import read_micromanager
+from iohub import read_images
 
 try:
     from pycromanager import Studio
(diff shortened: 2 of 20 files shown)


waveorder/acq/acq_functions.py
```diff
@@ -164,8 +164,8 @@ def acquire_from_settings(
         files = glob.glob(path + "*")
         index = max([int(x.split(path + "_")[1]) for x in files])
 
-        reader = read_micromanager(
+        reader = read_images(
             path + f"_{index}", data_type="ometiff", extract_data=True
         )
 
-        return reader.get_array(0)
+        return reader[0].compute()
```
talonchandler, line 171: Is this correct? I know you're unable to test pycromanager, but will this return an array?

tests/cli_tests/test_cli.py
```diff
@@ -7,5 +7,5 @@ def test_main():
     runner = CliRunner()
     result = runner.invoke(cli)
 
-    assert result.exit_code == 0
+    assert result.exit_code == 2
```
talonchandler, line 10: 2 seems like an error code. Is this expected?

waveorder/plugin/tab_recon.py
```diff
@@ -8,54 +8,31 @@
...
 import concurrent.futures
-import importlib.metadata
 
-from pydantic.v1 import BaseModel, NonNegativeInt
+from pydantic.v1 import BaseModel, NonNegativeInt, ValidationError
 
 from waveorder.cli import jobs_mgmt, settings
 from waveorder.io import utils
 
-try:
```
talonchandler, line 35: Great. Thanks @ziw-liu.

waveorder/plugin/tab_recon.py
```diff
@@ -2656,9 +2631,7 @@ def get_pydantic_kwargs(
         for field, field_def in pydantic_model.__fields__.items():
             if field_def is not None and field not in excludes:
                 ftype = field_def.type_
-                if isinstance(ftype, BaseModel) or isinstance(
-                    ftype, ModelMetaclass
-                ):
+                if isinstance(ftype, BaseModel):
```
talonchandler, line 2634: Are these changes needed?

tests/cli_tests/test_cli.py
ziw-liu: https://github.com/pallets/click/pull/1489

talonchandler (approved):
Thanks @ziw-liu. 
 
Following a discussion, I agree that this PR is necessary for us to move forward, and we do not need to rigorously test the `tab_recon.py` GUI right now. At the same time, do all of the changes in `tab_recon.py` file need to be included in this PR? I like the removal of the pydantic import checking, but I'm unsure about the other changes.

waveorder/acq/acq_functions.py
ziw-liu: Now using the MMStack reader directly so the LSP can infer the return type here.

waveorder/plugin/tab_recon.py
ziw-liu: In v2 this is not supported https://github.com/pydantic/pydantic/issues/6381. It's available in _some_ versions from a internal module but I don't understand this use case enough.

ziw-liu: And trying to import a non-existent symbol will prevent the entire plugin from starting.

ziw-liu: Hmm so it is available from the v1 namespace despite what maintainers said on that issue... reverted in https://github.com/mehta-lab/waveorder/pull/448/commits/b50820a59a6ad0d6dcc8ccda5134ccf77009c8d2

ziw-liu: I still can't get the recon tab to work though:

1. use a random HCS store that has a 'BF' channel as both input and output
2. click 'Load' and choose the example phase config
3. `KeyError: 'birefringence'`

talonchandler: Sounds good. I'll tag @amitabhverma here so he can see this, and let's move ahead with this merge for now.

Thanks @ziw-liu.

talonchandler (approved):

PR #457, opened 2025-07-08
Skip reconstructions of empty FOVs

ieivanov:
This PR skips the reconstruction of empty FOVs which results in NaNs in the zarr store and causes problems with downstream virtual staining


talonchandler (approved):
Thanks @ieivanov. 

I just benchmarked this check on a random (100, 2048, 2048) array and found that averaged ~0.5 seconds. 

Worth it in my opinion. Merging.

PR #469, opened 2025-08-17
Transform .md links to .html for sphinx docs

talonchandler:


amitabhverma:
> @amitabhverma I tested the docs locally and found this conversion to work well. Please check and use this if it solves your problem.

Did this solution work for `guide.html` which is generated using the `include` method, .md file in the .rst file ?
https://github.com/mehta-lab/waveorder/blob/0c9a2a7c000c2ec4f4ca1a3ebafbbd92ec580746/docs/sphinx_source/guide.rst?plain=1#L4-L5
It does work for .md files converted via the `myst_parser` directly (`napari-plugin-guide.html`) but does not seem to work for cases where its being included in a .rst.
I am using the README.md in `/guide` as the landing page as it is used on github, if an Index view of the landing page is acceptable then this method can be used. I've shared the previews below..

image

vs

image

talonchandler:
> Did this solution work for guide.html which is generated using the include method, .md file in the .rst file ?

Have you tried it locally? In my tests I was able to see the landing page and click the links from the guide.html. The links are broken on this link https://waveorder.readthedocs.io/en/latest/guide.html, but I don't have control (or understanding) of which branch is being displayed there.   

> I am using the README.md in /guide as the landing page as it is used on github, if an Index view of the landing page is acceptable then this method can be used. I've shared the previews below..

I prefer the landing page. The section navigation on the left seems to provide a top level view and the landing page gives a bit more guidance. 

Thanks for your help here @amitabhverma.

PR #473, opened 2025-09-02
Non-integer focus support

### LINKED ISSUE #470: `z_focus_offset` and `focus_from_transverse_band` do not support non-integer values
@gav-sturm's OPS applications will benefit from non-integer values of `z_focus_offset` and focus finding. This feature will improve our 2D phase reconstructions, especially when we have coarsely sampled slices. 

https://github.com/mehta-lab/waveorder/blob/0d3fe54e3123d2150b201a8683982a6ef7d517a4/waveorder/cli/settings.py#L65

talonchandler:
Fixes #470.

Enables sub-pixel precision for focus detection and `z_focus_offset` values, improving 2D phase reconstructions for coarsely sampled data. 

**Changes:**
- Settings: `z_focus_offset` now accepts float values (previously int only)
- Focus detection: Added optional sub-pixel precision to `focus_from_transverse_band()`
  - New parameter: `enable_subpixel_precision: bool = False`
  - Uses a polynomial fit to find a continuous maximum
  - Returns float indices when enabled, maintains integer behavior by default
- Backwards compatible. Default behavior return integers to continue to support mantis and other applications.

**Old usage (unchanged):**
```
focus_slice = focus_from_transverse_band(data, NA_det, lambda_ill, pixel_size)
```

**New usage:**
```
focus_slice = focus_from_transverse_band(
    data, NA_det, lambda_ill, pixel_size,
    polynomial_fit_order=2,
    enable_subpixel_precision=True
)  # Returns float like 5.3 instead of 5
```

**Config files now support float offsets:**
```
transfer_function:
  z_focus_offset: 0.5  # Previously only integers
```


gav-sturm:
@talonchandler one thing that might help in the documentation is selecting a reasonable polynomial fit order if subpixel_precission is set to True. I saw in your tests that 4 is a good default?

talonchandler:
> @talonchandler one thing that might help in the documentation is selecting a reasonable polynomial fit order if subpixel_precission is set to True. I saw in your tests that 4 is a good default?

Good suggestion. @gav-sturm, for your use case I suggest starting with order=2, then increasing to 3 or 4 if you're unhappy with the fits you're seeing in the plots. If this suggestion works well, I'll add it to the docstring.

srivarra (approved):
Looks good, I also tried messing with a torch and dask version, and didn't really see a big lift in performance.

PR #476, opened 2025-09-05
tests: run example tests on ci and mock show

tlambert03:
Hello!

Just starting to play around with waveorder...  This PR is just to see if we can successfully unskip the example tests on CI (and not have to manually close them all when running locally)

PR #477, opened 2025-09-05
run tests on pr

tlambert03:
adding a trigger to run tests on Pull requests

currently, tests only run on `push` ... which will work for all pushes to, say, branches... but *won't* run on a PR from a fork

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,6 +1,9 @@
 name: test
 
-on: [push]
+on:
+  push:
+    # branches: [main]  # uncomment if branch-PRs are causing dual triggers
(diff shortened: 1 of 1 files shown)


mattersoflight:
@talonchandler this PR came about after @tlambert03 updated tests so that examples run during CI (#476). Opening the PR #476 didn't trigger the tests, so this PR updates the github actions to do so. 

It looks like the CI will take ~20 mins. If you are okay with this configuration (CI on push and PR from forks), please approve and merge. 

We can review and merge #476 right after.

talonchandler (approved):
Thanks @tlambert03!

PR #478, opened 2025-09-05
simplify test ci

tlambert03:
i think the patterns in the test.yaml look largely borrowed from the napari cookiecutter suggestions, which, while aren't necessarily "outdated" per se, but can be simplified.  These PR exchanges three external github actions, for a single one (which i use in most of my qt-based CI tests these days)

I stopped short of removing tox... just in case you like to use it locally.  But you really aren't using tox's strengths here (which is that it can parametrize environment creation ... over more than just, say, OS and python... which is already being handled  by your github matrix).  If you want, I could remove that too


tlambert03:
> presumably unrelated to changes here?

but to be honest, I'm not 100% convinced of this.  It seems to consistently hang at around 47% on macos

talonchandler (approved):
Great thanks! Not sure on the slow macos tests either. I'm seeing the same thing on #476, so I think this branch is fine. Merging.

PR #489, opened 2025-09-10
Package layout fix

talonchandler:
Fixes a installation problem introduced by #480.

Specifying `[tool.setuptools.packages.find]; where = ["waveorder"]` looks for a package named `waveorder` inside `waveorder` which was one too deep...leading to import issues.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -118,8 +118,5 @@ line_length = 79
 [tool.setuptools_scm]
 write_to = "waveorder/_version.py"
 
-[tool.setuptools.packages.find]
-where = ["waveorder"]
-
 [tool.setuptools.package-data]
 "*" = ["*.yaml"]


srivarra (approved):
Very odd that GHA tests passed for that one, but locally just trying to use the `waveorder` cli and napari plugins failed.

srivarra:
As an aside, I did a quick test for converting the package from a [flat-layout][flat-doc] to a [src-layout][src-doc] and it worked flawlessly (tests, opening `napari` with `waveorder`, etc...). So if at any point you'd like to move to that layout structure it'll be smooth.

[flat-doc]: https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#flat-layout

[src-doc]: https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#src-layout

PR #492, opened 2025-09-12
Bump iohub

### LINKED ISSUE #491: Bump iohub
Start testing iohub 0.3 pre-releases.

talonchandler:
Fixes #491


ziw-liu:
I don't have any specific concerns. Maybe double check that CLI reconstruction with multi-processing still works?

PR #493, opened 2025-09-12
Hide acquisition tab

talonchandler:
This PR hides the acquisition tab, removes the relevant documentation section, and removes the documentation images.

There's a larger job of cleaning up `main_widget.py`, removing the `waveorder/acq` folder, etc, but it's not as important as this initial change.


waveorder/plugin/main_widget.py
```diff
@@ -100,6 +100,11 @@ def __init__(self, napari_viewer: Viewer):
         self.ui.setupUi(self)
         self.ui.tab_reconstruction.set_viewer(napari_viewer)
 
+        # Hide acquisition tab
```
srivarra: [question(non-blocking)](https://pullpo.io/cc?l=question&d=non-blocking) 
why not go all the way and remove that entire ui element?

srivarra (dismissed):

srivarra (approved):

PR #512, opened 2026-01-27
Backwards compatible stats from `focus_from_transverse_band`

talonchandler:
Replaces #179. 

`shrimPy` currently depends on `waveorder`'s [`return_peak_stats`](https://github.com/mehta-lab/waveorder/tree/return_peak_stats) branch, which introduces a breaking change to the `focus_from_transverse_band` API, which is used by the OPS project and others. `shrimPy` will need to depend on newer features of `waveorder` soon, so we'll need to sort this out. 

Here I'm proposed a backwards-compatible change to `focus_from_transverse_band`, where the `return_statistics` flag returns a statistics dictionary. This is the alternative the @ieivanov proposed at the start #179, instead of the more involved alternative I proposed. 

This PR is paired with [a `shrimPy` PR](https://github.com/czbiohub-sf/shrimPy/pull/223). I've put this change on a new branch instead of the existing `return_peak_stats` branch, so that this does not accidentally introduce a silent breaking change for `shrimPy`.

### DIFF
--- a/tests/test_focus_estimator.py
+++ b/tests/test_focus_estimator.py
@@ -156,3 +156,56 @@ def test_compute_midband_power_consistency():
 
     expected_focus_slice = np.argmax(manual_powers)
     assert focus_slice == expected_focus_slice
+
+
+def test_focus_from_transverse_band_with_statistics():
+    """Test focus_from_transverse_band with return_statistics=True."""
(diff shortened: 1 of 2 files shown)


ieivanov (approved):
Thanks @talonchandler!

PR #514, opened 2026-01-28
Update and test recon tab for pydantic v2

talonchandler:
#508 accidentally broke a couple pieces of the reconstruction tab. This PR fixes and tests the recon tab.

### DIFF
--- a/waveorder/plugin/tab_recon.py
+++ b/waveorder/plugin/tab_recon.py
@@ -2570,10 +2570,6 @@ class Ui_ReconTab_Form(QWidget):
                     self.add_pydantic_to_container(
                         unwrapped_ftype, new_widget, excludes, json_val
                     )
-                # ToDo: Implement Union check, tried:
-                # pydantic.typing.is_union(ftype)
-                # isinstance(ftype, types.UnionType)
-                # https://stackoverflow.com/questions/45957615/how-to-check-a-variable-against-union-type-during-runtime
                 elif isinstance(ftype, type(Union[NonNegativeInt, List, str])):
                     if (
                         field == "background_path"


waveorder/plugin/tab_recon.py
```diff
@@ -1309,19 +1402,18 @@ def build_model(self, selected_modes):
...
+                # Update the field
+                data[attr_key] = attr_val
+                # Recreate model from dict
+                model = settings.ReconstructionSettings.model_validate(data)
+
+            # Update the attribute directly
+            setattr(model, attr_key, attr_val)
         except Exception as exc:
             return print(exc.args)
         return model
```
srivarra: [issue(blocking)](https://pullpo.io/cc?l=issue&d=blocking) 
A couple of issues with `fix_model`:

1. `return print(exc.args)` returns `None` silently therefore the caller thinks it succeeded
2. Instead of looping over the `exclude_modes`, we can do `model.model_dump(exclude=set(exclude_modes))`.
3. `setattr` at the end is redundant as the value is already set from `model_validate`

waveorder/plugin/tab_recon.py
```diff
@@ -2478,29 +2574,51 @@ def add_pydantic_to_container(
...
+                    def_val = getattr(py_model, field)
+                else:
+                    def_val = field_def.default
+                # Pydantic v2: use annotation instead of type_
+                ftype = field_def.annotation
                 toolTip = ""
+                # Pydantic v2: class_validators is not available in FieldInfo
+                # Validators are now stored differently in v2
+                # For now, skip this since it's just for tooltips
+                # TODO: Update to use py_model.model_fields[field].metadata if needed
```
srivarra: I'll add this in.

waveorder/plugin/tab_recon.py
srivarra: I think we can simplify the logic here overall.

srivarra (approved):
Made a few adjustments, this is good to go.

Issue #503, opened 2025-11-07
Problem with the intallation of Waveorder CLI


JoseMMuino:
Dear All,

Thanks for your software Wavefront!
When I follow your instructions to install Wavefront, it doesnt install the CLI version.
When I run pip install waveorder[all] I get the warning "WARNING: waveorder 2.2.1 does not provide the extra 'all'"

When I run pip install waveorder=3.0.0a0, everything works fine and I can run waveorder CLI

When I try 
git clone https://github.com/mehta-lab/waveorder.git
cd waveorder
pip install -e .[all]

I get:
[...]
Preparing metadata (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
  × Preparing metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [31 lines of output]
      Traceback (most recent call last):
        File "/home/muino/miniforge3/envs/waveorder/lib/python3.14/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 389, in <module>
          main()
          ~~~~^^
        File "/home/muino/miniforge3/envs/waveorder/lib/python3.14/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 373, in main
          json_out["return_val"] = hook(**hook_input["kwargs"])
                                   ~~~~^^^^^^^^^^^^^^^^^^^^^^^^
        File "/home/muino/miniforge3/envs/waveorder/lib/python3.14/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 178, in prepare_metadata_for_build_wheel
          whl_basename = backend.build_wheel(metadata_directory, config_settings)
        File "/tmp/pip-build-env-8yztbef7/overlay/lib/python3.14/site-packages/sipbuild/api.py", line 28, in build_wheel
          project = AbstractProject.bootstrap('wheel',
                  arguments=_convert_config_settings(config_settings))
        File "/tmp/pip-build-env-8yztbef7/overlay/lib/python3.14/site-packages/sipbuild/abstract_project.py", line 74, in bootstrap
          project.setup(pyproject, tool, tool_description)
          ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/tmp/pip-build-env-8yztbef7/overlay/lib/python3.14/site-packages/sipbuild/project.py", line 661, in setup
          self.apply_user_defaults(tool)
[...]

PR #517, opened 2026-02-03
Constrain `PyQt6>=6.5,<6.8` for older linux systems

talonchandler:
Biohub's HPC depends on Rocky Linux 8 -> glibc 2.28-2.33 -> manylinux_2_28 -> PyQt6 6.5-6.8. If we don't constrain `PyQt6>=6.5,<6.8`, PyQt6 wheels are unavailable and PyQt6 will try to use `qmake` to build PyQt6 which typically fails. 

Related: 
- https://github.com/mehta-lab/recOrder/pull/503
- https://github.com/mehta-lab/waveorder/issues/503
- https://github.com/napari/napari/issues/7976

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,6 +64,7 @@ dynamic = ["version"]
 [project.optional-dependencies]
 visual = [
   "napari[pyqt6]",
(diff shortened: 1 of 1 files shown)


talonchandler:
> Do we know if this closes those related issues?

Not 100% sure...I'll leave those issues open for now.

PR #522, opened 2026-02-12
Batched focused slice

edyoshikun:
Paired with https://github.com/mehta-lab/VisCy/pull/364

- allow for batched T*Z GPU compute
- From the little benchmarking on a small node using it at 60-75% VRAM capacity it sped up by 3x. This function itself doesn't automatically handle the batching but `viscy/preprocessing/focus.py` in viscy does.


talonchandler (approved):
LGTM!

PR #527, opened 2026-02-26
Adds support for Micro-Manager acquired ome tif

amitabhverma:
- Added CLI support for Micro-Manager acquired ome tif datasets: converts to .zarr and then proceeds to reconstruction.
- Added GUI functionality to support ome-tif
- Added metadata info next to dataset selector utilizing "iohub info" like functionality

### DIFF
--- a/waveorder/cli/utils.py
+++ b/waveorder/cli/utils.py
@@ -243,7 +243,10 @@ def check_folder_for_ometiff(input_data_folder: Path) -> bool:
     """
     Checks for Micro-Manager ome-tif folder
     """
-    data_type, extra_info = _infer_format(input_data_folder)
+    try:
+        data_type, extra_info = _infer_format(input_data_folder)
+    except (ValueError, RuntimeError):
+        return False
     if data_type == "ometiff":
         return True
     return False


waveorder/cli/reconstruct.py
```diff
@@ -43,8 +46,16 @@ def _reconstruct_cli(
     See /examples for example configuration files.
 
     >> waveorder reconstruct -i ./input.zarr/*/*/* -c ./examples/birefringence.yml -o ./output.zarr
```
talonchandler: I would suggest this example string should show something like `./input_tif_or.zarr/*/*/*`

waveorder/cli/utils.py
```diff
@@ -190,3 +195,140 @@ def _check_nan_n_zeros(input_array):
...
+    try:
+        data_type, extra_info = _infer_format(input_data_folder)
+    except (ValueError, RuntimeError):
+        return False
+    if data_type == "ometiff":
+        return True
+    return False
+
+
+def get_dataset_info(path: str):
```
talonchandler: My understanding is that this is for populating the tooltip string in the GUI. 

Is there any reason this information can't be used to autopopulate GUI fields?

PR #528, opened 2026-03-06
Optimizing blind reconstruction #525

JoOkuma:
Most tests are passing, except those in `tests/test_examples.py`, because I don't have the files.
A (9, 410, 410) tile for 10 iterations runs in ~1.4 seconds  
For 50 iterations, it takes about ~2.9 seconds
Benchmark was run on my local 3090, it should be faster in the cluster.


talonchandler:
Closing...see #532 (PR), #533, and #534 (potential optimizations). Thanks @JoOkuma

PR #532, opened 2026-03-07
Reconstruction optimization and GPU support

talonchandler:
Supersedes #528 and the computation speedup portions of #526 (both had stale targets). The CUDA streaming pipeline from #526 will follow as a separate PR from @srivarra.

**Contributions**

This PR integrates optimizations from @jookuma (#528) and @mark-a-potts (#526), rebased onto current main and adapted to work with the API layer from #525.

- @jookuma: batched WOTF, batched `nd_fourier_central_cuboid`, `stretched_matrix_multiply`, `gen_coordinate` → torch, `torch.no_grad()` for grid quantities, z-index pre-allocation
- @mark-a-potts: verbose print removal, device placement pattern

**Speedup per commit at (9, 410, 410)**

| Commit | Iter time | vs main |
|--------|-----------|---------|
| **main** (baseline) | 270ms | 1.0x |
| batch WOTF + fourier cuboid | 208ms | 1.3x |
| batch filter bank (`stretched_matrix_multiply`) | 159ms | 1.7x |
| `gen_coordinate` → torch (fftfreq) | 150ms | 1.8x |
| `torch.no_grad()` for grid quantities | ~150ms | 1.8x |
| NullLogger (silent by default) | ~145ms | 1.9x |
| `device="mps"` (Apple Silicon) | **80ms** | **3.4x** |

Reconstruction outputs are unchanged, all diffs are floating-point noise.

### DIFF
--- a/docs/examples/demos/QLIPP/QLIPP_simulation.py
+++ b/docs/examples/demos/QLIPP/QLIPP_simulation.py
@@ -123,7 +123,9 @@ jupyter_visuals.plot_multicolumn(
 xx, yy, fxx, fyy = util.gen_coordinate((N, M), ps)
 radial_frequencies = torch.sqrt(fxx**2 + fyy**2)
 Source_cont = optics.generate_pupil(radial_frequencies, NA_illu, lambda_illu).numpy()
-Source_discrete = optics.Source_subsample(Source_cont, lambda_illu * fxx, lambda_illu * fyy, subsampled_NA=0.1)
+Source_discrete = optics.Source_subsample(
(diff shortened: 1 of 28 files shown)


JoOkuma (commented):
Hi @talonchandler , thanks for reporting the individual improvements of each change.

I left a few comments. I tried to suggest the changes through GitHub, so there might be some typos and linting problems.

I think we could improve the performance a bit by generating the grids only once (e.g., `gen_coordinate`), but the code might become a bit less functional.

For example, there could be a bit `ComputeMidbandPower` class that stores the grid for the loss function.

JoOkuma (approved):

srivarra (approved):
Very exciting changes!

PR #554, opened 2026-04-23
Replace create_empty_hcs_zarr with iohub's create_empty_plate

ieivanov:
## Summary
- Remove `waveorder.cli.utils.create_empty_hcs_zarr`. iohub's `create_empty_plate` does the same plate/channel creation work.
- The only waveorder-specific behavior was writing a `plate_metadata` dict into the plate's zattrs; that's done inline in `apply_inverse_transfer_function.py` after `create_empty_plate` returns, matching the pattern used in biahub (czbiohub-sf/biahub#215, czbiohub-sf/biahub#241).
- Port the two tests that referenced the helper (`tests/util_tests/test_create_empty.py`, `tests/cli_tests/test_zarr_v3_reconstruct.py`) to `create_empty_plate`.

## Why not also replace `apply_inverse_to_zyx_and_save`?
iohub's `process_single_position` is numpy-native: it loads `ndarray` slices via `oindex` and writes them back the same way. `apply_inverse_to_zyx_and_save` operates on `xr.DataArray` — it selects channels by name, passes the xarray to the waveorder reconstruction functions (`birefringence/phase/fluorescence.apply_inverse_transfer_function`, all xarray-in/xarray-out), and writes via `output_position.write_xarray(...)` so the `t` coord and attributes are preserved. A drop-in swap would either require refactoring the reconstruction API to become numpy-native (loses xarray coord semantics and breaks public API) or adding a numpy↔xarray wrapper that duplicates logic iohub doesn't expose. Deferred.

## Test plan
- [x] `uv run pytest` — 347 passed, 1 skipped, 2 xfailed.

### DIFF
--- a/tests/cli_tests/test_zarr_v3_reconstruct.py
+++ b/tests/cli_tests/test_zarr_v3_reconstruct.py
@@ -21,7 +21,7 @@ from waveorder.cli.apply_inverse_transfer_function import (
     get_reconstruction_output_metadata,
 )
 from waveorder.cli.main import cli
-from waveorder.cli.utils import create_empty_hcs_zarr
+from iohub.ngff.utils import create_empty_plate
 from waveorder.io import utils
 
 INPUT_SCALE = [1, 1, 2.0, 6.5, 6.5]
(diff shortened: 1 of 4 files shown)


talonchandler (approved):
Thanks @ieivanov!

PR #558, opened 2026-04-29
Use ProcessPoolExecutor with spawn for apply-inv-tf workers

ieivanov:
## Summary

Replace `torch.multiprocessing.Pool().starmap()` with `ProcessPoolExecutor(mp_context=mp.get_context("spawn"))` in `apply_inverse_transfer_function_single_position`. Mirrors the pattern already used in `iohub.ngff.utils.process_single_position`.

Two motivations:

- **spawn (not fork)** — tensorstore runs internal C++ threads that are not fork-safe; a forked worker can deadlock or segfault before our code runs (see google/tensorstore#61).
- **`ProcessPoolExecutor` (not `mp.Pool`)** — a silently dying worker (e.g. cgroup OOM-kill) surfaces as `BrokenProcessPool` instead of hanging indefinitely on `pool.starmap`.

`-j` continues to flow through unchanged: `num_processes` becomes `max_workers`. The pool tops out at `len(time_indices)` workers (one task per timepoint), same as before.

## Test plan

- [x] Smoke-tested on a real HCS plate (8 positions, T=10, C=3, Z=86, Y=1600, X=1370) with `biahub reconstruct -j 16` on SLURM. All 6 SLURM jobs completed; outputs are bit-equivalent to the prior `mp.Pool` version (same logs, same data).
- [x] Verified inside a running job that 10 worker processes (capped by T=10) are spawned under the parent and consume all 16 allocated cores.
- [x] Verified `num_processes <= 1` path still runs serially in the calling process.
- [x] Reviewer to confirm CI passes.

### DIFF
--- a/waveorder/cli/apply_inverse_transfer_function.py
+++ b/waveorder/cli/apply_inverse_transfer_function.py
@@ -367,13 +367,8 @@ def apply_inverse_transfer_function_single_position(
         # (e.g. cgroup OOM-kill) surfaces as BrokenProcessPool instead
         # of hanging indefinitely on pool.starmap.
         context = mp.get_context("spawn")
-        with ProcessPoolExecutor(
-            max_workers=num_processes, mp_context=context
-        ) as p:
-            futures = [
-                p.submit(partial_apply_inverse_to_zyx_and_save, t_idx)
-                for t_idx in time_indices
-            ]
+        with ProcessPoolExecutor(max_workers=num_processes, mp_context=context) as p:
+            futures = [p.submit(partial_apply_inverse_to_zyx_and_save, t_idx) for t_idx in time_indices]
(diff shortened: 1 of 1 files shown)


ieivanov:
@talonchandler this change matches what we use in `iohub` `process_single_position`: https://github.com/czbiohub-sf/iohub/blob/1c739a21ae6370aaaef22b23343d4a13bbb9fbce/src/iohub/ngff/utils.py#L449-L457. Before this change, `biahub reconstruct` job would stall when using multiple workers. It would be great if you can merge and create a new release

srivarra (approved):

talonchandler:
Thanks! Merged and released: https://github.com/mehta-lab/waveorder/releases/tag/3.0.4

PR #559, opened 2026-05-05
Tile Stitch Engine

srivarra:
Signed-off-by: Sricharan Reddy Varra <sricharan.varra@biohub.org>


talonchandler:
My commit changes `tile-stitch` to wrap the entire reconstruction .yaml, not just the `phase/fluorescecence` part. This changes enables all of the features within waveorder reconstruct (channel selection, 2D/3D reconstruction), and removes the need for a `kind` field in the `tile-stitch` config.

talonchandler:
With the change above this PR now doesn't change the existing configs, so I'm confident it's safe to merge and iterate with. Thanks for doing this work @srivarra.

PR #564, opened 2026-06-19
MM Ome tif support

amitabhverma:
- Added CLI support for Micro-Manager acquired ome tif datasets: converts to .zarr and then proceeds to reconstruction.
- Added GUI functionality to support ome-tif
- Added metadata info next to dataset selector utilizing "iohub info" like functionality

Incorporated changes based on comments from previous pull request #527

### DIFF
--- a/docs/guide/napari-plugin-guide.md
+++ b/docs/guide/napari-plugin-guide.md
@@ -86,11 +86,11 @@ It is normal to see background retardance and orientation. We will use these bac
 The advanced tab gives the user a log output which can be useful for debugging purposes. There is a log level “debugging” which serves as a verbose output. Look here for any hints as to what may have gone wrong during calibration or acquisition.
 
 ## Reconstruction Tab
-The **Reconstruction** tab is designed to reconstruct `birefriengence, phase, birefrignence with phase, and flurescenece` datasets that have been either acquired or coverted to `.zarr` store as well as acquisitions that are in progress.
+The **Reconstruction** tab is designed to reconstruct `birefriengence, phase, birefrignence with phase, and flurescenece` datasets that have been either acquired as Micro-Manager `.ome-tif` or coverted to `.zarr` store as well as acquisitions that are in progress.
 
 ![](../_static/images/reconstruction_data.png)
 
-The **Input Store** and **Output Directory** point to the input and output `.zarr` data locations. Once an Input Store is selected some metadata parameters can be viewed by hovering the cursor over the `info label` ⓘ.
+The **Input Store** and **Output Directory** point to the input and output (`.ome-tif`,`.zarr`) data locations. Once an Input Store is selected some metadata parameters can be viewed by hovering the cursor over the `info label` ⓘ.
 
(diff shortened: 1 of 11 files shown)


talonchandler (changes requested):
Hi @amitabhverma, thanks for this iteration.

The main shape of this feature is correct, but much of what's here duplicates code that lives in `iohub`. I just pushed a [`feature/ometiff-input` branch](https://github.com/mehta-lab/waveorder/tree/feature/ometiff-input) that refactors to show what a minimal version would look like. By reusing `iohub` and moving the switching in `parsing.py`, we reduce the feature diff to <50 lines. 

Regarding pinned dependencies, let's figure this out with automated tests. In my example branch I added a test that checks a simple ome.tiff against the current pyproject.toml, and I couldn't reproduce the issues you mentioned. If you saw failures, can you share:

- the traceback,
- the iohub, tifffile, zarr versions in the failing environment,
- the path/dataset that triggered the issue, and
- ideally, a failing automated test that reproduces the problem. 

If we need to pin versions, this PR should include a change to pyproject.toml and an automated test that fails without the pin.

talonchandler (dismissed):
Final fixes made together. Looks great! Thanks @amitabhverma

amitabhverma:
@talonchandler 

Couple of points I forgot to ask/share on Friday.

1. For ome-zarr metadata [print_info](https://github.com/czbiohub-sf/iohub/blob/28c2a1f3cb08163895f3bb21706bbc48d6214c53/src/iohub/reader.py#L201) does not print out the Scale and FOV which are the 2 additions I had to do on my end. Do you use something else to get that info using iohub ? Ideally, like we discussed, this function should reside in iohub with a wrapper on waveorder.
 
image


2. I moved some other metadata parsing which was under tab_recon to [utils.get_dataset_info](https://github.com/mehta-lab/waveorder/blob/9afca388b26acfea7d658d2280e218a10a63ba8c/waveorder/cli/utils.py#L263) which adds those extra fields and also provides easier access to fields that might be needed for auto-populate. Regarding auto-populate, is it explicit that there would only be a single channel when doing a Phase/Fluorescence reconstruction ?

3. I was trying the pytest and trying to locate what you had shown me and realized the version on https://waveorder.readthedocs.io/en/latest/guide/development-guide.html is different from whats in the [/main branch docs](https://github.com/mehta-lab/waveorder/blob/main/docs/guide/development-guide.md) are. I believe I had it set to automatically generate the html docs with a new commit but it seems like it's not done that for the last 10 months.
4. I got the tests running locally now but had to manually install `pytest-qt`

talonchandler:
Thanks Amit...following up with a summary of our chat just now:

1. I'm flexible on this...you can leave it as is and/or pass in single FOVs in the print_info. 

2. Yes, only a single channel for fluorescence and phase. Birefringence recons can take multiple input channels.

3. No worries. Github docs is fine for now...Sri will likely move + rebuild our docs later. 

4. We worked through this together...see CONTRIBUTING.md. Feel free to flag/fix conflicting instructions.

PR #566, opened 2026-06-24
fix: replace ProcessPoolExecutor with ThreadPoolExecutor (#552)

### LINKED ISSUE #552: Replace mp.Pool with ThreadPoolExecutor in apply_inverse_transfer_function_single_position
## Summary

`apply_inverse_transfer_function_single_position` uses `multiprocessing.Pool` to parallelize reconstruction across timepoints. In pipeline contexts (e.g. Nextflow + Slurm), the orchestrator already handles parallelism at the position level — each Slurm job processes a single position. Within that job, `mp.Pool` forks child processes that each get a full copy of the input data (via fork + copy-on-write), leading to very high memory usage and poor performance on large volumes.

Since the core compute (numpy/scipy 3D FFTs, Tikhonov solves) releases the GIL, thread-based parallelism would avoid the data-copying overhead while achieving equivalent CPU utilization.

## Observed behavior

- **Input**: 5 timepoints × 1 channel × 86 Z × 1600 Y × 1370 X, float32 (~750 MB per timepoint)
- **Settings**: 3D Tikhonov with `TV_iterations: 1`, `z_padding: 5`, `num_processes=4`
- **Context**: Nextflow dispatches one Slurm job per position (4 positions running concurrently), each calling `apply_inverse_transfer_function_single_position` with `num_processes=4`
- **Result**: 0 out of 5 timepoints completed after 1h16m. Each of the 4 worker processes consumed 35–45 GB RSS (vs ~750 MB per volume), indicating massive memory duplication from forking.

## Relevant code

https://github.com/mehta-lab/waveorder/blob/main/waveorder/cli/apply_inverse_transfer_function.py

```python
with mp.Pool(num_processes) as p:
    p.starmap(...)
```

## Suggestion

Rather than replacing the existing `mp.Pool` path (which may work well for standalone CLI usage), it would be useful to add an alternative thread-based parallelization option — e.g. a `num_threads` parameter that uses `concurrent.futures.ThreadPoolExecutor` instead.

This is analogous to what iohub did in [czbiohub-sf/iohub#396](https://github.com/czbiohub-sf/iohub/pull/396) (landed in iohub v0.3.1), where `process_single_position` was switched from `multiprocessing.Pool` to `ThreadPoolExecutor` since numpy/scipy/torch all release the GIL.

Thread-based parallelism avoids the memory duplication that makes `mp.Pool` expensive when an external orchestrator (Nextflow, Snakemake, etc.) is already handling coarse-grained parallelism across positions.

## Environment

- waveorder 3.0.2
- Python 3.12
- HPC node: 16 CPUs, 244 GB RAM
- Lustre filesystem

parallelArchitect:
Fixes #552

## Problem

Process-based parallelism duplicates zarr-backed data into each worker
via copy-on-write. On large volumes this causes severe memory inflation —
#552 reports 35–45 GB RSS for a 750 MB dataset with 4 workers.

Two additional hazards with the process-based approach:

1. **tensorstore fork-safety**: tensorstore runs internal C++ threads
   that are not fork-safe. A forked worker inherits locked mutexes from
   threads that don't exist in the child, causing deadlocks or segfaults
   (google/tensorstore#61). The `spawn` context avoided forking but added
   significant startup overhead per worker.

2. **PyTorch thread limiting**: `torch.set_num_threads(1)` and
   `set_num_interop_threads(1)` were required to prevent thread explosion
   across spawned processes. This artificially constrained PyTorch's
   internal parallelism.

## Fix

Replace `ProcessPoolExecutor` with `ThreadPoolExecutor`. Threads share
memory — no copying, no fork-safety hazards, no spawn overhead.

numpy, scipy, and PyTorch all release the GIL during compute (FFTs,
Tikhonov solves), so threads achieve genuine parallelism on the
compute-heavy parts of the reconstruction pipeline.

## Changes

- `waveorder/cli/parsing.py`: rename `processes_option` → `threads_option`,
  replace `--num_processes` with `--num-threads`. Deprecated aliases
  `--num-processes` / `--num_processes` retained for backward compatibility.
- `waveorder/cli/apply_inverse_transfer_function.py`: swap executor,
  remove spawn context and torch thread limiting.
- `waveorder/cli/reconstruct.py`: update to use `threads_option` and
  `num_threads`.
- `waveorder/cli/utils.py`: rename `num_processes` → `num_threads` in
  `estimate_resources`.
- `waveorder/calib/calibration_workers.py`: update keyword argument.

## Testing

200 tests pass locally (cli_tests, api_tests, models).
`waveorder/waveorder_simulator.py` uses `ProcessPoolExecutor` for a
different purpose (CPU-bound Jones matrix computation, no zarr I/O) and
is intentionally left unchanged.

### DIFF
--- a/waveorder/calib/calibration_workers.py
+++ b/waveorder/calib/calibration_workers.py
@@ -334,7 +334,7 @@ class BackgroundCaptureWorker(CalibrationWorkerBase, signals=BackgroundSignals):
             transfer_function_dirpath=transfer_function_path,
             config_filepath=reconstruction_config_path,
             output_dirpath=reconstruction_path,
-            num_processes=1,
+            num_threads=1,
         )
 
         # Load reconstructions from file for layers
--- a/waveorder/cli/apply_inverse_transfer_function.py
+++ b/waveorder/cli/apply_inverse_transfer_function.py
@@ -9,7 +9,7 @@ from waveorder.cli.parsing import (
     config_filepath,
     input_position_dirpaths,
     output_dirpath,
-    processes_option,
+    threads_option,
     transfer_function_dirpath,
     write_config_scale_to_output,
 )
@@ -203,17 +203,16 @@ def apply_inverse_transfer_function_single_position(
     transfer_function_dirpath: Path,
     config_filepath: Path,
     output_position_dirpath: Path,
-    num_processes,
+    num_threads,
     output_channel_names: list[str],
     verbose: bool = True,
 ) -> None:
 
     # Deferred imports for fast CLI help
-    from concurrent.futures import ProcessPoolExecutor, as_completed
+    from concurrent.futures import ThreadPoolExecutor, as_completed
     from functools import partial
 
     import numpy as np
-    import torch.multiprocessing as mp
     from iohub import open_ome_zarr
 
     from waveorder.api import (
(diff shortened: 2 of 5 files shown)


ieivanov:
The PoolExecutor has been very carefully chosen to work with our biahub reconstruction pipeline. I advise not changing the default. We could offer an option to slot in a different executor if needed for certain applications.

PR #568, opened 2026-06-25
feat: anisotropic yx pixel size for fluorescence and phase

### LINKED ISSUE #567: Support anisotropic yx pixel sizes in fluorescence and phase reconstructions
waveorder assumes square pixels (single `yx_pixel_size` scalar). DaXi (and sometimes mantis) produce data where `y != x`, and we want to reconstruct it without resampling first.

**Scope.** Fluorescence (2D + 3D) and phase (2D + 3D) get full anisotropic support. 

**Config (backwards compatible).**

```yaml
yx_pixel_size: 0.3

yx_pixel_size:
  y: 0.3
  x: 0.25
```

The scalar form parses to an isotropic value internally, so existing configs are unchanged.

**De-risking.** Anisotropic call with `y == x` produces identical transfer functions and reconstructions to the legacy scalar call. New anisotropic tests cover each modality. Regression benchmark suite still passes.

Sri will validate on DaXi zebrafish data before merge.

talonchandler:
Closes #567.

Adds support for non-square pixels in fluorescence (2D + 3D) and phase (2D + 3D) reconstructions. Birefringence paths raise `NotImplementedError` when `y != x`. Scalar configs and `y == x` configs keep working everywhere.

**Config (backwards compatible).**

```yaml
yx_pixel_size: 0.3

yx_pixel_size:
  y: 0.3
  x: 0.25
```

Both forms parse to a normalized `YXPixelSize(y, x)` internally.

**What changed.** New `YXPixelSize` value type with a `from_value` constructor that accepts scalar, mapping, or itself. Pydantic settings field flipped to `YXPixelSize` with a before-validator and a serializer that emits a scalar when isotropic so YAML stays tidy. `util` helpers, all phase and fluorescence models, the `api/phase` and `api/fluorescence` consumers, and the CLI output metadata path all accept `YXPixelSize | float`. Anisotropic Nyquist upsampling (separate `y_factor` and `x_factor`) in the four affected models. One new anisotropic example at `docs/examples/api/phase_3d_anisotropic.py`.

**Tests (de-risking).** 46 new tests. Settings: scalar and dict YAML parsing, YAML round-trip in both forms, typo and negative rejection. Utilities: isotropic equivalence vs legacy scalar, anisotropic frequency and sphere generation. Each modality (phase 2D/3D, fluor 2D/3D): isotropic equivalence (`y == x` reproduces legacy output bit for bit) plus anisotropic smoke. Birefringence: anisotropic input raises `NotImplementedError`. CLI: anisotropic config writes distinct y and x scales to the output zarr, mismatch warning reports y and x independently. Full regression benchmark suite passes with identical metrics.

**Out of scope.** Birefringence anisotropic support. Shift-variant fluorescence. Z pixel size stays scalar.

Sri to validate on DaXi zebrafish data before this merges.

### DIFF
--- a/waveorder/_pixel_size.py
+++ b/waveorder/_pixel_size.py
@@ -26,8 +26,8 @@ class YXPixelSize(BaseModel):
 
     model_config = ConfigDict(extra="forbid", frozen=True)
 
-    y: PositiveFloat = Field(description="pixel size along y in micrometers")
-    x: PositiveFloat = Field(description="pixel size along x in micrometers")
+    y: PositiveFloat = Field(default=0.1, description="pixel size along y in micrometers")
+    x: PositiveFloat = Field(default=0.1, description="pixel size along x in micrometers")
 
     @classmethod
     def isotropic(cls, value: float) -> YXPixelSize:
--- a/waveorder/_pixel_size.py
+++ b/waveorder/_pixel_size.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 from typing import Any
 
-from pydantic import BaseModel, ConfigDict, Field, PositiveFloat
+from pydantic import BaseModel, ConfigDict, Field, PositiveFloat, model_serializer
 
 
 class YXPixelSize(BaseModel):
@@ -77,3 +77,14 @@ class YXPixelSize(BaseModel):
     def is_isotropic(self) -> bool:
         """True if y and x spacings are equal."""
         return self.y == self.x
+
+    @model_serializer
+    def _serialize(self):
+        """Serialize as a scalar when isotropic, else as ``{'y': ..., 'x': ...}``.
+
+        Both forms round-trip through :meth:`from_value`, so YAML configs
+        for square pixels stay readable as a single number.
+        """
+        if self.y == self.x:
+            return self.y
+        return {"y": self.y, "x": self.x}
--- a/waveorder/_pixel_size.py
+++ b/waveorder/_pixel_size.py
@@ -1,0 +1,79 @@
+"""Lateral pixel size value type, supporting isotropic and anisotropic yx."""
+
+from __future__ import annotations
+
+from typing import Any
+
+from pydantic import BaseModel, ConfigDict, Field, PositiveFloat
+
+
+class YXPixelSize(BaseModel):
+    """Lateral pixel size in micrometers, possibly anisotropic.
+
+    Use :meth:`isotropic` for square pixels, or :meth:`from_value` to
+    normalize either a scalar, a mapping, or an existing YXPixelSize
+    into a :class:`YXPixelSize`. The ``from_value`` constructor is the
+    intended entry point for internal functions that accept either a
+    plain float (treated as isotropic) or this model.
+
+    Attributes
+    ----------
+    y : float
+        Pixel size along the y axis in micrometers.
+    x : float
+        Pixel size along the x axis in micrometers.
+    """
+
+    model_config = ConfigDict(extra="forbid", frozen=True)
+
+    y: PositiveFloat = Field(description="pixel size along y in micrometers")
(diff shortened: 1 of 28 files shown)


talonchandler:
Hi @srivarra, give this a try on your next zebrafish reconstruction trial.

srivarra:
@talonchandler Cool, I will give it a shot.

talonchandler:
Ready to go from my perspective. Thanks for trying this out @srivarra.

srivarra (approved):

srivarra (changes requested):
Gonna unapprove real quick, accidentally clicked that button.

waveorder/_pixel_size.py
```diff
@@ -0,0 +1,101 @@
...
+    y : float
+        Pixel size along the y axis in micrometers.
+    x : float
+        Pixel size along the x axis in micrometers.
+    """
+
+    model_config = ConfigDict(extra="forbid", frozen=True)
+
+    y: PositiveFloat = Field(default=0.1, description="pixel size along y in micrometers")
+    x: PositiveFloat = Field(default=0.1, description="pixel size along x in micrometers")
```
srivarra: If a user passes only `y` or only `x`, the other axis silently takes the default of `0.1` without feedback to the user. We should try to validate that the mapping contains both keys or uses the isotropic version.


We also probably want some guards here:

```python
y: PositiveFloat = Field(..., allow_inf_nan=False)
x: PositiveFloat = Field(..., allow_inf_nan=False)
```

waveorder/_pixel_size.py
```diff
@@ -0,0 +1,101 @@
...
+        # 0-d ndarrays, 0-d tensors. ``float()`` raises TypeError for
+        # multi-element arrays and non-numeric inputs.
+        try:
+            as_float = float(value)
+        except (TypeError, ValueError):
+            raise TypeError(
+                f"cannot convert {type(value).__name__} to YXPixelSize; "
+                "expected float, mapping with keys 'y' and 'x', or YXPixelSize"
+            ) from None
+        return cls.isotropic(as_float)
```
srivarra, line 103: Pydantic can't convert a `TypeError` to `ValidationError`, but `ValueError` and `AssertionError` are understood by Pydantic. We should raise `ValueError` instead of `TypeError` here.

waveorder/_pixel_size.py
```diff
@@ -0,0 +1,101 @@
...
+        return self.y == self.x
+
+    @model_serializer
+    def _serialize(self):
+        """Serialize as a scalar when isotropic, else as ``{'y': ..., 'x': ...}``.
+
+        Both forms round-trip through :meth:`from_value`, so YAML configs
+        for square pixels stay readable as a single number.
+        """
+        if self.y == self.x:
```
srivarra: since we have the `is_isotropic` method, we should use that instead of rechecking again.

srivarra (changes requested):
One thing I saw that's not touched here (not even sure if it needs to be modified) is that in `optics.py` 

https://github.com/mehta-lab/waveorder/blob/d3de793f9e79d9b9404725e8caf818c820fb79df/waveorder/optics.py#L194-L199

`df` comes from `fxx` only, but `pixel_slope` sets the sigmoid roll-off in both directions. With anisotropic pixels `df_y != df_x`.

Do we need `df_y` here too?

talonchandler:
Thanks @srivarra, all of these make sense...do you mind quickly taking care of them?

>df comes from fxx only, but pixel_slope sets the sigmoid roll-off in both directions. With anisotropic pixels df_y != df_x.

> Do we need df_y here too?

Let's do `df = torch.min(torch.abs(fxx[0, 1] - fxx[0, 0]), torch.abs(fyy[1, 0] - fyy[0, 0]))`.

srivarra (approved):
good to go!

PR #573, opened 2026-08-03
Add RL/RLGC deconvolution for 3D fluorescence

talonchandler:
Adds Richardson-Lucy (RL) and Gradient-Consensus (RLGC) deconvolution for 3D fluorescence.

Select it in a config with `reconstruction_algorithm: RL` or `RLGC`, and tune `rl_iterations`, `rl_background`, and `rl_stopping_tolerance`. RLGC resists the noise overfitting ("starry night") that RL shows when over-iterated.

RL/RLGC options raise `NotImplementedError` for 2D fluorescence and for phase/birefringence.

The operator-agnostic solver lives in `waveorder/rlgc.py`, adapted from Andrew York's Gradient Consensus demo ([doi.org/10.5281/zenodo.10278918](https://zenodo.org/records/10278919)). Tests in `tests/models/test_rlgc.py` cover the adjoint operators, bead sharpening under Poisson noise, RL vs RLGC over-iteration, and the `NotImplemented` guards.

### DIFF
--- a/docs/examples/cli/configs/fluorescence_2d.yml
+++ b/docs/examples/cli/configs/fluorescence_2d.yml
@@ -15,7 +15,10 @@ fluorescence:
     wavelength_emission: 0.532              # emission wavelength in micrometers
     confocal_pinhole_diameter: null         # confocal pinhole diameter (null = widefield)
   apply_inverse:
-    reconstruction_algorithm: Tikhonov      # 'Tikhonov' or 'TV' regularization
+    reconstruction_algorithm: Tikhonov      # 'Tikhonov'/'TV' filters or 'RL'/'RLGC' iterative deconvolution
     regularization_strength: 0.01           # strength of regularization
     TV_rho_strength: 0.001                  # ADMM rho parameter for TV regularization
     TV_iterations: 1                        # ADMM iterations for TV regularization
+    rl_iterations: 25                       # maximum RL / RLGC iterations
+    rl_background: 0.0                      # constant background folded into the RL / RLGC Poisson forward model
+    rl_stopping_tolerance: null             # relative-change early-stop threshold for RL / RLGC (null = run all iterations)
--- a/docs/examples/cli/configs/fluorescence_2d.yml
+++ b/docs/examples/cli/configs/fluorescence_2d.yml
@@ -19,11 +19,4 @@ fluorescence:
     regularization_strength: 0.01           # strength of regularization
     TV_rho_strength: 0.001                  # ADMM rho parameter for TV regularization
     TV_iterations: 1                        # ADMM iterations for TV regularization
-    rl_iterations: 25                       # maximum RL / RLGC iterations
-    rl_background: 0.0                      # constant background folded into the RL / RLGC Poisson forward model
-    rl_stopping_tolerance: null             # relative-change early-stop threshold for RL / RLGC (null = run all iterations)
-    rl_back_projector: matched              # 'matched' is the matched transpose (classic RL); the unmatched 'gaussian'/'butterworth'/'wiener'/'wiener_butterworth' converge in far fewer iterations but are supported for 'RL' only, not 'RLGC'
-    rl_bp_alpha: null                       # Wiener regularization for the 'wiener'/'wiener_butterworth' back projectors (null = matched cutoff gain)
-    rl_bp_beta: null                        # cutoff gain for the 'butterworth'/'wiener_butterworth' back projectors (null = matched cutoff gain)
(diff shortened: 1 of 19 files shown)


srivarra (dismissed):
I think this looks good to me. Do we want to wait to merge, or just go ahead?

docs/examples/cli/configs/fluorescence_2d.yml
```diff
@@ -15,7 +15,15 @@ fluorescence:
...
     TV_iterations: 1                        # ADMM iterations for TV regularization
+    rl_iterations: 25                       # maximum RL / RLGC iterations
+    rl_background: 0.0                      # constant background folded into the RL / RLGC Poisson forward model
+    rl_stopping_tolerance: null             # relative-change early-stop threshold for RL / RLGC (null = run all iterations)
+    rl_back_projector: matched              # 'matched' is the matched transpose (classic RL); the unmatched 'gaussian'/'butterworth'/'wiener'/'wiener_butterworth' converge in far fewer iterations but are supported for 'RL' only, not 'RLGC'
+    rl_bp_alpha: null                       # Wiener regularization for the 'wiener'/'wiener_butterworth' back projectors (null = matched cutoff gain)
+    rl_bp_beta: null                        # cutoff gain for the 'butterworth'/'wiener_butterworth' back projectors (null = matched cutoff gain)
+    rl_bp_order: 8                          # Butterworth order for the 'butterworth'/'wiener_butterworth' back projectors
+    rl_bp_resolution_mode: fwhm             # cutoff-frequency rule for the back projector ('fwhm_over_sqrt2' suits iSIM)
```
talonchandler, line 29: @srivarra can we think of a way to clean up these configs a bit? I'm okay if there are lots of parameters in the lower-level methods, but I'd prefer to keep the user-facing configs as clean as we can. If we're not careful this list will get very long very fast. 

- is there a reasonable way to nest these parameters? For example, regularization strength only applies if 'Tikhonov' mode is being used. `rl` parameters only apply if `RL` or `RLGC` is being used. 
- Have you found it useful to play with `rl_bp_order`? If not, you can leave it off of the config for now. 
- I'm not sure with `rl_bp_resolution_mode` does. I'm guessing you can drop it for now. 

Thanks!

talonchandler (commented):
Thanks, left one comment to clean up the parametrization a bit. Otherwise I think we're close to ready.

docs/examples/cli/configs/fluorescence_2d.yml
srivarra: yeah they're pretty messy. I think a solution is to create another Pydantic settings object specifically for RL, and we can set it to be optional for the fluorescence reconstruction yaml configuration.

I'm going to play with this and let you know if it's useful or not. It's the exponent of the butterworth and so we'll just leave it out.

`rl_bp_resolution_mode` helps the butterworth and weiener-butterworth know where the resolution limit is. It's measured from the OTF, but you can pick the rule for tuning the measurement into a cutoff. It's either the full width at half maximum or that divided by sqrt(2). Either way, I think we can drop this parameter as well for the config.

srivarra:
@talonchandler 

Here's an example of an updated yaml config for RL:

```yaml 
input_channel_names: [GFP]
time_indices: all
reconstruction_dimension: 3
fluorescence:
  transfer_function:
    yx_pixel_size: 0.1                  # scalar, or {y: 1.018, x: 0.1842} for anisotropic
    z_pixel_size: 0.25
    z_padding: 0
    index_of_refraction_media: 1.3
    numerical_aperture_detection: 1.2
    wavelength_emission: 0.532
    confocal_pinhole_diameter: null     # null = widefield
  apply_inverse:
    reconstruction_algorithm: RL
    rl:
      iterations: 100
      back_projector: matched
      background: 0.0
      stopping_tolerance: null
```

and if you want to use Tikhonov:

```yaml
input_channel_names: [GFP]
time_indices: all
reconstruction_dimension: 3
fluorescence:
  transfer_function:
    yx_pixel_size: 0.1
    z_pixel_size: 0.25
    z_padding: 0
    index_of_refraction_media: 1.3
    numerical_aperture_detection: 1.2
    wavelength_emission: 0.532
    confocal_pinhole_diameter: null
  apply_inverse:
    reconstruction_algorithm: Tikhonov
    regularization_strength: 0.001
```
So if a user runs RL they have another block of parameters to set, but if they don't, it doesn't need to be filled in.

talonchandler:
LGTM! Thanks @srivarra...made a couple minor changes, and I think it's ready for final review and merge.

srivarra (approved):

PR #576, opened 2026-08-12
fix: replace `torch.tensor` copy-constructs with `as_tensor`

### LINKED ISSUE #139: Torch warning about copying tensors
`recOrder` tests surfaced the following warnings at `optics.py:321`, `optics.py:370`, and `stokes.py:291`
```
recOrder/tests/cli_tests/test_reconstruct.py::test_reconstruct
  /Users/talon.chandler/waveorder/waveorder/optics.py:321: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).
    * torch.tensor(z_position_list)[:, None, None]
```

You can generate these warnings with a call like
```
import torch
torch.tensor(torch.tensor([0,1,2]))[:,None,None]
```

The core issue seems to be that some `optics` and `stokes` functions can take either a list or a tensor. 

@ziw-liu do you think we should make sure that only lists are passed? Or can you think of a simpler solution?

This does not seem critical.

srivarra:
clears up some warnings, and removes the double copy if you do `torch.tensor(x).clone()` and stops the autograd graph from detaching with `torch.tensor()`.

### DIFF
--- a/tests/test_stokes.py
+++ b/tests/test_stokes.py
@@ -130,19 +130,6 @@ def test_copying(device):
     assert a[0] == 1
 
 
-@pytest.mark.parametrize(*_DEVICE)
-def test_estimate_copying(device):
-    s = torch.tensor([1.0, 1.0], device=device)
-
-    _, _, transmittance, _ = stokes.estimate_adr_from_stokes(s, s, s, s)
-    transmittance[0] = 2  # modify the output
-    assert s[0] == 1  # check that the input hasn't changed
-
-    _, _, transmittance012 = stokes.estimate_ar_from_stokes012(s, s, s)
-    transmittance012[0] = 2
-    assert s[0] == 1
-
-
 def test_gradients_reach_copied_outputs():
     """s0 and transmittance are copies of an input, so they must not detach."""
     ones = torch.ones((2, 2))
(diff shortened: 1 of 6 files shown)


ieivanov:
LGTM overall. One optional follow-up: the sweep leaves a few conversions behind in the files it already touches.

- `waveorder/optim/losses.py:178` (the midband-power loss, just above the spectral-flatness one this PR fixes) still has `frr = np.sqrt(fxx**2 + fyy**2)`. `gen_coordinate` returns tensors, so this goes through NumPy's `__array_wrap__` and emits the NumPy 2 deprecation warning. `torch.sqrt(fxx**2 + fyy**2)` would match the change at the spectral-flatness loss.
- `waveorder_reconstructor.py` (544, 546, 584, 586, 594, 596, 776, 785, 907, 909, 1245, 1247, 1262–1265) and `waveorder_simulator.py` (113, 115) still call `torch.tensor(...)` on NumPy arrays in the same calls that now use `torch.as_tensor(self.frr)` / `torch.as_tensor(self.Source)`, e.g. `torch.tensor(self.Pupil_support)`, `torch.tensor(z)`. They're harmless copies, but it's inconsistent within the same call. None of the receiving functions modify their inputs in place, so `as_tensor` would be safe there too.

The two `torch.tensor([...])` list literals in `stokes.py` (78, 90) are fine as they are.

ieivanov (approved):

PR #175, opened 2024-09-10
Wrap-safe transfer functions

talonchandler:
Fixes #152. Fixes #172. 

This PR implements wrap-safe transfer functions. After this PR, undersampled data will not create wrapped transfer functions, which should address many of the reconstruction artifacts we've seen. 

Before (e.g. phase transfer function with transverse and axial wrapping):
Screenshot 2024-09-09 at 8 10 36 PM

After:
Screenshot 2024-09-09 at 8 11 14 PM

A prerequisite for this PR is a clear set of [transverse and axial Nyquist sampling rates](https://czbiohub.atlassian.net/wiki/spaces/TLG3Infect/pages/3671457799/2024-06-21+what+is+the+axial+Nyquist+sampling+rate). 

This PR redefines the `calculate_transfer_function` call so it should Just Work (™). The old calls have been renamed to `_calculate_wrap_unsafe_transfer_function`. Better naming suggestions are welcome. 

This PR implements these fixes for all of the transfer functions in the main branch: 
- [x] Widefield fluorescence
- [x] 3D phase 
- [x] 2D phase

Similar behavior for vector-optics transfer functions will follow in the `vector-prototype` branch. 

**Testing status** I'm passing automated tests, and I've tested all of the example scripts with undersampled parameters to check for wrapping. @ziw-liu I would appreciate your help with testing on real data because your use cases helped us spot these artifacts.

PR #577, opened 2026-08-31
Add high-frequency apodization knob to reduce checkerboarding artifacts

talonchandler:
Screenshot 2026-08-31 at 3 54 04 PM

We've observed high-frequency ringing in 2D reconstructions from OPS and other datasets, colloquially "checkerboarding". These artifacts are often large enough to be visually distracting (particularly with tight contrast limits), and they might mask valuable features.

I've isolated this issue to reconstructions of datasets that are sampled slightly below Nyquist, where our anti-wrapping efforts (#175) cause a sharp edge in frequency space. 

Potential solutions are:

1. Reconstruct onto a finer grid then downsample. This lets us avoid all wrapping issues, but it increases memory and compute requirements by >4x for OPS datasets.
2. Add a high-frequency apodization filter knob, to enable tuning of the reconstruction's response near the sampling Nyquist frequency.

Here I'm suggesting option 2 with a single parameter, `apodization_rolloff`, on `apply_inverse_transfer_function` and `reconstruct`. When > 0, the inverse filter is multiplied by a separable raised-cosine window that is 1 over the inner `1 - rolloff` fraction of the band and rolls smoothly to exactly zero at the transverse Nyquist frequency, so the reconstruction contains no energy at the high frequencies that create the checkerboard. Low- and mid-band transfer is unchanged.

**Details**

- Default is 0.0, which preserves current behavior.
- Applies identically to labelfree and fluorescence reconstructions.
- Increasing Tikhonov regularization suppresses small singular values, so it is different than this knob. For label-free imaging, small singular values can be found at high **and* low frequencies, while this knob only reduces high frequencies. 
- Recommended value of 0.25 for OPS.

### DIFF
--- a/waveorder/sampling.py
+++ b/waveorder/sampling.py
@@ -67,6 +67,35 @@ def axial_nyquist(
     return 1 / (2 * cutoff_frequency)
 
 
+def raised_cosine_window(size, rolloff, device=None):
+    """1D raised-cosine apodization window in unshifted (fftfreq) order.
+
+    The window is 1 over the inner ``1 - rolloff`` fraction of the band
+    and rolls off smoothly (half-cosine) to 0 at the Nyquist frequency.
+
+    Parameters
+    ----------
+    size : int
+        Number of frequency samples.
+    rolloff : float
+        Fraction of the band (0 to 1] over which the window rolls off
+        from 1 to 0 at the band edge.
+    device : str, torch.device, or None
+        Output tensor device.
+
+    Returns
+    -------
+    torch.Tensor
+        Window of shape ``(size,)`` in fftfreq (unshifted) order.
+    """
+    normalized_frequency = torch.abs(torch.fft.fftfreq(size, device=device)) * 2  # 1.0 at Nyquist
+    rolloff_start = 1 - rolloff
+    window = torch.ones(size, device=device)
+    mask = normalized_frequency > rolloff_start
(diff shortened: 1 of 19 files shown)


waveorder/sampling.py
```diff
@@ -67,6 +67,35 @@ def axial_nyquist(
...
+    device : str, torch.device, or None
+        Output tensor device.
+
+    Returns
+    -------
+    torch.Tensor
+        Window of shape ``(size,)`` in fftfreq (unshifted) order.
+    """
+    normalized_frequency = torch.abs(torch.fft.fftfreq(size, device=device)) * 2  # 1.0 at Nyquist
+    rolloff_start = 1 - rolloff
```
srivarra, line 92: What ranges of values for `rolloff` are reasonable? I'm wondering b/c if we use something like `1e-8` `rolloff_start` just rounds to 1.

srivarra (changes requested):
Let's make sure to expose `apodization_rolloff` in the config settings because `MyBaseModel` has forbids extra fields that arn't there in the model class. Looks good otherwise!

gav-sturm:
I'll give this a shot on Wednesday mostly to check for easy application of the fix on OPS.

talonchandler:
Thanks @srivarra @gav-sturm.

@srivarra I just threaded this through the CLI and improved the documentation. Thanks!

gav-sturm:
looks great! claude had no trouble using the new parameter in ops_process: paired PR is here: https://github.com/royerlab/ops_process/pull/185

srivarra (approved):
LGTM 👍

PR #578, opened 2026-09-04
feat: choose regularization_strength from the data (otsu_cnr, l_curve)

ArianaGu:
Add an optional `auto_regularization` block to `apply_inverse`. When a config carries one, `wo rec` sweeps `regularization_strength` after computing the transfer function, picks a value by one of two rules, writes it into a sibling `<config>_autoreg.yml`, and reconstructs from that. There is no new CLI flag: the block is the trigger, as an `lr` key is for `optimization:`.

Try it:

    cd docs/examples/cli
    bash phase_3d_autoreg.sh

which simulates a phantom and reconstructs it with the sweep on, using `configs/phase_3d_autoreg.yml`, and prints:

    Choosing phase regularization_strength with the 'otsu_cnr' rule...
        regularization_strength = 1.129e-07
        lambda / |H|^2max       = 0.0001
        sweep index             = 2 of 9
        scored crop             = 256x256 at (y=0, x=0), t=0

Alongside the reconstruction it leaves two artifacts: `configs/phase_3d_autoreg_autoreg.yml`, the config with the pick frozen in and the block spent, so rerunning it reproduces the reconstruction without sweeping again; and `autoreg_report.json`, every strength, score and norm from the sweep. The output zarr is bit-identical to reconstructing with that lambda written by hand, since auto-regularization only resolves the config and then hands it to the unchanged reconstruction path.

To enable it on an existing config:

    apply_inverse:
      reconstruction_algorithm: Tikhonov
      regularization_strength: 0.001          # the sweep overwrites this
      auto_regularization:
        rule: otsu_cnr
        search_range: [-6.0, 2.0]
        num_samples: 25
        search_scale: relative
        report_path: ./autoreg_report.json

The options
-----------

- `rule` — which criterion picks the value. `otsu_cnr` (default) maximizes contrast-to-noise of an Otsu split of the local-variance map, scored on one shared z slice so the sweep stays comparable. `l_curve` takes the corner of the residual/solution trade-off, from curvature on range-normalized log axes. Both are ported from the CZ Biohub weight-search scripts.

- `search_range` — log10 bounds of the sweep, read according to `search_scale`. Default `[-6.0, 2.0]`.

- `num_samples` — how many strengths to try across that range, evenly spaced in log10. At least 3, since `l_curve`'s curvature is a second derivative. Default 25.

- `search_scale` — `relative` (default) reads `search_range` as log10(lambda / |H|^2max); `absolute` reads it as log10(lambda). Relative is the default because lambda only means something against |H|^2max, which sets how far the filter attenuates its own peak, and that value spans ~6 decades between modalities (1.3e-6 for a brightfield phase CTF, 1.0 for a normalized fluorescence OTF), so absolute decades do not port between configs. |H|^2max is set by the optics and the pixel size rather than the frame size, which is what lets the sweep score a 256x256 crop and still emit a strength suited to the full frame.

- `report_path` — where to write the sweep as JSON: every strength with its score or its two norms, the chosen index, lambda/|H|^2max, the scored crop, and any warnings. `null` (default) writes nothing.

Everything else the sweep needs is fixed rather than exposed: the 256x256 max-variance scoring crop, the local-variance window, and the timepoint, which is the first entry of the config's own `time_indices`.

The sweep rebuilds the inverse filter rather than calling the model per value, so it applies `apodization_rolloff` the same way the models do; a test on an undersampled configuration, where the window actually bites, pins that.

`l_curve` is why this is a sweep rather than a use of the existing `optimization:` machinery: its criterion is a property of the whole trace of norms, not of any single reconstruction, so it cannot be phrased as a loss.

Scope
-----

3D Tikhonov: `phase:` or `fluorescence:` without `birefringence:`. Three cases drop the block while parsing, each with a visible
`AutoRegularizationIgnoredWarning`: thin (2D) reconstructions, which regularize a singular system rather than a single filter; joint birefringence+phase, which shares one `regularization_strength` between the phase filter and the vector singular system; and non-Tikhonov algorithms. These warn rather than raise because the napari plugin unwraps `Optional[Model]` fields and always submits a populated block (`plugin/tab_recon.py::get_pydantic_kwargs`), so rejecting on a block's presence would break every 2D reconstruction started from the GUI. The warning has its own `filterwarnings` carve-out, as `PixelSizeMismatchWarning` does, because `cli/main.py` suppresses `UserWarning`.

Neither rule is an oracle
-------------------------

The module docstring says so too. On real brightfield phase data, L-curves from a band-limited operator are frequently not L-shaped (the residual floors well above zero), and the `otsu_cnr` pick moves with the scoring crop. So a pick landing on a sweep endpoint warns, the crop and lambda/|H|^2max are always echoed, and `report_path` exists for review.

Verification
------------

A swept reconstruction at lambda_i is bit-identical to `apply_inverse_transfer_function(regularization_strength=lambda_i)` for both models, with and without z padding, so the sweep scores the production reconstruction. Both rule ports are pinned against their sources: `otsu_cnr` agrees with `batched_otsu_cnr` to 0.000e+00 on a recorded fixture and on real 2048x2048 data, and `find_l_curve_corner` reproduces the reference index on L-shaped, near-linear and noisy curves. The crop-versus-full-frame |H|^2max claim above was measured at 0.000000 decades on a 2048x2048 frame and 0.005 decades across 128/256/512/1024 for both models. On real data a 2048x2048x16 fluorescence reconstruction takes 18.4 s with the sweep against 16.9 s without, at unchanged peak memory.

Adds 19 unit tests (30 parametrized cases), 8 settings tests and a CLI smoke test alongside `test_optimization_cli`. `phase_3d_autoreg.yml` is generated from the settings model by `test_generate_example_settings`, so it cannot drift, and `phase_3d_autoreg.sh` runs in CI with the other CLI examples.

Note: `wo rec` cannot read big-endian (`>u2`) zarrs, as the zarrs codec pipeline raises "Non-native byte order not supported", with or without this change. Unrelated, but worth an issue.

### DIFF
--- a/docs/examples/cli/configs/phase_3d_autoreg.yml
+++ b/docs/examples/cli/configs/phase_3d_autoreg.yml
@@ -21,4 +21,10 @@ phase:
     TV_rho_strength: 0.001                  # ADMM rho parameter for TV regularization
     TV_iterations: 1                        # ADMM iterations for TV regularization
     apodization_rolloff: 0.0                # raised-cosine roll-off fraction applied to the inverse filter at the transverse Nyquist edge; suppresses checkerboard artifacts when the optical band limit exceeds the sampling Nyquist frequency. Must be between 0 and 1 (0 = off); if you see checkerboarding artifacts, start with 0.25
-    auto_regularization: null               # choose regularization_strength from the data; 3D Tikhonov only
+    auto_regularization:                    # choose regularization_strength from the data; 3D Tikhonov only
+      rule: otsu_cnr                        # 'otsu_cnr' maximizes local-variance contrast; 'l_curve' finds the trade-off corner
+      search_min: -6.0                      # lower log10 bound of the sweep, interpreted per search_scale
+      search_max: 2.0                       # upper log10 bound of the sweep, interpreted per search_scale
+      num_samples: 9                        # regularization strengths to try
+      search_scale: relative                # 'relative' reads the search bounds as log10(lambda / |H|^2max); 'absolute' as log10(lambda)
+      report_path: ./autoreg_report.json    # write the full sweep (scores, norms, chosen index) to this JSON path; null = no report
--- a/docs/examples/cli/phase_3d_autoreg.sh
+++ b/docs/examples/cli/phase_3d_autoreg.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+# Reconstruct 3D phase with `regularization_strength` chosen from the data.
+#
+# `wo rec` sweeps the strength after computing the transfer function, writes the
+# value it picked into `phase_3d_autoreg_autoreg.yml`, and dumps the whole sweep
+# to `autoreg_report.json`. Read the report before trusting the pick: neither
+# rule is an oracle, and `wo rec` warns when the pick lands on a sweep endpoint.
+wo sim \
+  -c ./configs/phase_3d_autoreg.yml \
+  -o ./phase_autoreg_data.zarr
+
+wo rec \
+  -i ./phase_autoreg_data.zarr \
+  -c ./configs/phase_3d_autoreg.yml \
+  -o ./phase_3d_autoreg_recon.zarr
+
+wo view ./phase_autoreg_data.zarr ./phase_3d_autoreg_recon.zarr
--- a/tests/optim/test_autoreg.py
+++ b/tests/optim/test_autoreg.py
@@ -1,0 +1,408 @@
+"""Tests for automatic regularization-strength selection.
+
+The reference numbers in this file were produced by the CZ Biohub weight-search
+scripts that :mod:`waveorder.optim.autoreg` ports (``compMicro/utils.py`` and
+``compMicro/weight_search.py``), so a change in behaviour shows up as a failure
+here rather than as a quietly different pick.
+"""
+
+import json
+
+import numpy as np
+import pytest
+import torch
+from pydantic import ValidationError
+
+from waveorder.models import isotropic_fluorescent_thick_3d, phase_thick_3d
+from waveorder.optim import autoreg
+from waveorder.optim.autoreg import AutoRegularizationSettings
+
+PHASE_TF_KWARGS = dict(
+    yx_pixel_size=0.15,
+    z_pixel_size=1.0,
+    wavelength_illumination=0.441,
+    index_of_refraction_media=1.3,
+    numerical_aperture_illumination=1.15,
+    numerical_aperture_detection=0.52,
+)
+FLUOR_TF_KWARGS = dict(
+    yx_pixel_size=0.325,
+    z_pixel_size=1.0,
+    wavelength_emission=0.525,
+    index_of_refraction_media=1.0,
+    numerical_aperture_detection=0.55,
+)
+
+
+def _measurement(zyx_shape=(12, 64, 64), seed=0):
+    rng = np.random.default_rng(seed)
+    return torch.from_numpy((rng.random(zyx_shape) * 100 + 500).astype(np.float32))
+
+
+def _phase_tfs(zyx_shape, z_padding):
+    return phase_thick_3d.calculate_transfer_function(zyx_shape=zyx_shape, z_padding=z_padding, **PHASE_TF_KWARGS)
+
+
+def _fluor_tf(zyx_shape, z_padding):
+    return isotropic_fluorescent_thick_3d.calculate_transfer_function(
+        zyx_shape=zyx_shape, z_padding=z_padding, **FLUOR_TF_KWARGS
+    )
+
+
+def _sweep_one(data, tf, contrast, z_padding, strength, apodization_rolloff=0.0):
+    """Reproduce a single point of the sweep, the way select_regularization does."""
+    measurement = autoreg.preprocess_measurement(data, contrast, z_padding)
+    spectrum = torch.fft.fftn(measurement, dim=(-3, -2, -1))
+    recon = torch.real(
+        torch.fft.ifftn(
+            spectrum * autoreg._inverse_filter(tf, strength, apodization_rolloff),
+            dim=(-3, -2, -1),
+        )
+    )
+    return recon[z_padding : recon.shape[0] - z_padding] if z_padding else recon
+
+
+# --- the load-bearing test: the sweep scores the production reconstruction ---
+
+
+@pytest.mark.parametrize("z_padding", [0, 3])
+@pytest.mark.parametrize("strength", [1e-8, 1e-3])
+def test_sweep_reconstruction_matches_phase_model(z_padding, strength):
+    """A swept reconstruction is what phase_thick_3d would have produced."""
+    zyx_shape = (12, 64, 64)
+    data = _measurement(zyx_shape)
+    real_tf, imag_tf = _phase_tfs(zyx_shape, z_padding)
+
+    expected = phase_thick_3d.apply_inverse_transfer_function(
+        data,
+        real_tf,
+        imag_tf,
+        z_padding=z_padding,
+        regularization_strength=strength,
+    )
+    actual = _sweep_one(data, real_tf, "phase", z_padding, strength)
+
+    assert torch.allclose(expected, actual, atol=1e-6)
+
+
+@pytest.mark.parametrize("z_padding", [0, 3])
+@pytest.mark.parametrize("strength", [1e-8, 1e-3])
+def test_sweep_reconstruction_matches_fluorescence_model(z_padding, strength):
+    """A swept reconstruction is what isotropic_fluorescent_thick_3d would have produced."""
+    zyx_shape = (12, 64, 64)
+    data = _measurement(zyx_shape)
+    otf = _fluor_tf(zyx_shape, z_padding)
+
+    expected = isotropic_fluorescent_thick_3d.apply_inverse_transfer_function(
+        data,
+        otf,
+        z_padding=z_padding,
+        regularization_strength=strength,
+    )
+    actual = _sweep_one(data, otf, "fluorescence", z_padding, strength)
+
+    assert torch.allclose(expected, actual, atol=1e-6)
+
+
+def test_sweep_applies_apodization_like_the_model():
+    """apodization_rolloff reshapes the inverse filter, so the sweep must apply it too.
+
+    Uses a pixel size that undersamples the optical band, which is the case the
+    knob exists for. At a well-sampled pixel size the window is ~1 wherever the
+    filter carries energy and this test would pass with or without the fix.
+    """
+    zyx_shape, rolloff = (12, 64, 64), 0.25
+    data = _measurement(zyx_shape)
+    undersampled = {**PHASE_TF_KWARGS, "yx_pixel_size": 0.5}
+    real_tf, imag_tf = phase_thick_3d.calculate_transfer_function(zyx_shape=zyx_shape, z_padding=0, **undersampled)
+
+    expected = phase_thick_3d.apply_inverse_transfer_function(
+        data, real_tf, imag_tf, z_padding=0, regularization_strength=1e-3, apodization_rolloff=rolloff
+    )
+    unapodized = _sweep_one(data, real_tf, "phase", 0, 1e-3)
+    apodized = _sweep_one(data, real_tf, "phase", 0, 1e-3, rolloff)
+
+    # The window has to make a real difference here, or the assertion below is empty.
+    assert (expected - unapodized).abs().max() > 0.05 * (expected.max() - expected.min())
+    assert torch.allclose(expected, apodized, atol=1e-6)
+
+
+def test_preprocess_measurement_normalizes_phase_only():
+    """Phase intensity-normalizes before its inverse filter; fluorescence does not."""
+    data = _measurement((6, 16, 16))
+
+    phase = autoreg.preprocess_measurement(data, "phase", z_padding=0)
+    fluorescence = autoreg.preprocess_measurement(data, "fluorescence", z_padding=0)
+
+    assert torch.allclose(phase, data / data.mean() - 1)
+    assert torch.allclose(fluorescence, data)
+
+    padded = autoreg.preprocess_measurement(data, "fluorescence", z_padding=2)
+    assert padded.shape == (10, 16, 16)
+
+    with pytest.raises(ValueError, match="contrast"):
+        autoreg.preprocess_measurement(data, "birefringence", z_padding=0)
+
+
+# --- rules, against reference values from the scripts they port ---
+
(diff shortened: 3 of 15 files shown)


docs/examples/cli/configs/phase_3d_autoreg.yml
srivarra, line 1: I think we should be able to consolidate the autoreg config specifics into the regular `phase_3d.yml` config file.

docs/examples/cli/phase_3d_autoreg.sh
srivarra, line 1: I think we can remove this file, do you think it's useful for others?

waveorder/optim/autoreg.py
```diff
@@ -0,0 +1,656 @@
...
+# Box-filter window for the local-variance map that otsu_cnr thresholds.
+OTSU_WINDOW_SIZE = 3
+
+_OTSU_NUM_BINS = 256
+_LOCAL_VAR_CLIP_QUANTILE = 0.99
+# torch.quantile rejects inputs beyond ~16M elements, which a full-frame volume
+# exceeds; sample down to this many values instead.
+_QUANTILE_MAX_ELEMENTS = 2**24
+_L_CURVE_PROMINENCE_FRACTION = 0.10
+_L_CURVE_PLATEAU_FRACTION = 0.90
```
srivarra, line 71: > Side length of the square yx crop that is scored

Does autoreg extend to anisotropic voxels? We sometimes have datasets that have different Y and X sizes.

waveorder/api/_settings.py
```diff
@@ -143,6 +147,27 @@ class FourierApplyInverseSettings(MyBaseModel):
         "limit exceeds the sampling Nyquist frequency. Must be between 0 and 1 (0 = off); "
         "if you see checkerboarding artifacts, start with 0.25",
     )
+    auto_regularization: Optional[AutoRegularizationSettings] = Field(
+        default=None,
+        description="choose regularization_strength from the data; 3D Tikhonov only",
+    )
```
srivarra, line 153: I think this exposes something we haven't had to handle yet in the existing GUI code. Even if the config has `auto_regularization: null`, the Napari plugin in `waveorder/plugin/tab_recon.py` turns it into a populated block when collecting the form values. For 3D Tikhonov reconstructions, that would trigger a sweep and overwrite the manual strength despite auto-regularization being disabled in the config. 

Could we update [`add_pydantic_to_container`](https://github.com/mehta-lab/waveorder/blob/49a35f5950924d4c70635725ec9d8f0e11f4d03b/waveorder/plugin/tab_recon.py#L2043-L2182) and [`get_pydantic_kwargs`](https://github.com/mehta-lab/waveorder/blob/49a35f5950924d4c70635725ec9d8f0e11f4d03b/waveorder/plugin/tab_recon.py#L2185-L2229) to preserve that disabled state unless the user explicitly enables it? A config loaded with `auto_regularization: null` should still submit `None` if the user leaves it untouched.

waveorder/optim/autoreg.py
```diff
@@ -0,0 +1,656 @@
...
+            "lambda_over_h2max": self.lambda_over_h2max,
+            "transfer_function_peak_squared": self.transfer_function_peak_squared,
+            "crop": {"y0": self.crop[0], "x0": self.crop[1], "size": self.crop[2]},
+            "regularization_strengths": _list(self.regularization_strengths),
+            "scores": _list(self.scores),
+            "residual_norms": _list(self.residual_norms),
+            "solution_norms": _list(self.solution_norms),
+            "warnings": list(self.warnings),
+        }
+
```
srivarra, line 187: [nitpick(if-minor)](https://pullpo.io/cc?l=nitpick&d=if-minor) 
I'm not sure if we need the `_list` helper function, we should just be able to convert it directly like so:

```python
"regularization_strengths": self.regularization_strengths.tolist(),
"scores": self.scores.tolist(),
"residual_norms": self.residual_norms.tolist(),
"solution_norms": self.solution_norms.tolist(),
"warnings": self.warnings.tolist(),
```

Since they're all already NumPy arrays (is that going to be the case?).

waveorder/optim/autoreg.py
```diff
@@ -0,0 +1,656 @@
...
+        lambda_over_h2max=strength / h2max,
+        crop=crop,
+        scores=scores,
+        residual_norms=residual_norms,
+        solution_norms=solution_norms,
+        warnings=messages,
+    )
+
+
+def warn_all(result: AutoRegResult) -> None:
```
srivarra, line 653: [issue(if-minor)](https://pullpo.io/cc?l=issue&d=if-minor) 
Looks like we already exempt `AutoRegularizationIgnoredWarning` from being suppressed by the CLI at the top of the file, but `warn_all` still emits a `UserWarning` so it'll be hidden. Could we give selection warnings the same treatment? 

We could do a separate category or generalize the `AutoRegularizationIgnoredWarning`. Up to you!

waveorder/optim/autoreg.py
```diff
@@ -0,0 +1,656 @@
...
+    """
+    inverse_filter = tikhonov_regularized_inverse_filter(transfer_function, strength)
+    if apodization_rolloff > 0:
+        window = sampling.raised_cosine_window(
+            inverse_filter.shape[-2], apodization_rolloff, device=inverse_filter.device
+        )[:, None] * sampling.raised_cosine_window(
+            inverse_filter.shape[-1], apodization_rolloff, device=inverse_filter.device
+        )
+        inverse_filter = inverse_filter * window
+    return inverse_filter
```
srivarra, line 505: [thought(non-blocking)](https://pullpo.io/cc?l=thought&d=non-blocking) 
We now repeat this filter construction and apodization block in the sweep and both 3D reconstruction models. We could fold the apodization into `tikhonov_regularized_inverse_filter` with an optional `apodization_rolloff` parameter, so all three callers share the same implementation. Nonblocking for this PR. I’ll likely handle this in a follow-up.

Here are the other places the `apodization_rolloff` occurs in:
- [autoreg.py:497–504](https://github.com/mehta-lab/waveorder/blob/49a35f5950924d4c70635725ec9d8f0e11f4d03b/waveorder/optim/autoreg.py#L497-L504)
 - [phase_thick_3d.py:439–447](https://github.com/mehta-lab/waveorder/blob/49a35f5950924d4c70635725ec9d8f0e11f4d03b/waveorder/models/phase_thick_3d.py#L439-L447)
 - [isotropic_fluorescent_thick_3d.py:361–369](https://github.com/mehta-lab/waveorder/blob/49a35f5950924d4c70635725ec9d8f0e11f4d03b/waveorder/models/isotropic_fluorescent_thick_3d.py#L361-L369)

waveorder/optim/autoreg.py
```diff
@@ -0,0 +1,656 @@
...
+    """
+    padded = util.pad_zyx_along_z(zyx_data, z_padding)
+    if contrast == "phase":
+        return util.inten_normalization_3D(padded)
+    if contrast == "fluorescence":
+        return padded
+    raise ValueError(f"contrast must be 'phase' or 'fluorescence', got {contrast!r}")
+
+
+def select_regularization(
```
srivarra, line 542: [question(if-minor)](https://pullpo.io/cc?l=question&d=if-minor) 
Could we compute the metrics for each reconstructed volume and then discard it, rather than retain the full stack? 

For L-curve, we only need the residual and solution norms per strength. Otsu needs a shared z slice across the sweep, but we could retain the per-slice selection statistics and CNR scores, then choose that slice afterward. That would require scoring every slice, but avoid keeping all the volumes in memory. 

The current stack grows with both the number of strengths and the Z depth, which could become a problem for datasets with 2,000+ slices (Zebrafish). Even with a 256 × 256 crop and 25 strengths, the reconstruction stack alone is about 13 GB. Some napkin math puts peak memory in the tens of GB, potentially approaching 100 GB once FFT buffers and temporary arrays are included. Processing one candidate at a time should bring that down substantially.

Let me know what you think about this one, b/c I'm not too sure.

tests/optim/test_autoreg.py
```diff
@@ -0,0 +1,408 @@
...
+    real_tf, imag_tf = _phase_tfs(zyx_shape, z_padding)
+
+    expected = phase_thick_3d.apply_inverse_transfer_function(
+        data,
+        real_tf,
+        imag_tf,
+        z_padding=z_padding,
+        regularization_strength=strength,
+    )
+    actual = _sweep_one(data, real_tf, "phase", z_padding, strength)
```
srivarra, line 83: [issue(if-minor)](https://pullpo.io/cc?l=issue&d=if-minor) 
Could we extend this equivalence check to call `select_regularization` directly? `_sweep_one` uses the preprocessing and filter helpers, so this already covers those well, but it assembles the reconstruction separately from the actual sweep.

For example if the sweep stopped passing through `apodization_rolloff` the helper-based test would still pass.

Maybe we could have a separate test, or having this one use `select_regularization`, up to you.

srivarra (changes requested):
Overall, I think this looks great. I'm excited to try out some reconstructions with Zebrafish data with the autoreg setup!

There's just a few small nitpicks more or less.

srivarra:
@ArianaGu 
> Note: wo rec cannot read big-endian (>u2) zarrs, as the zarrs codec pipeline raises "Non-native byte order not supported", with or without this change. Unrelated, but worth an issue.

Huh, how'd you run into this issue?

waveorder/optim/autoreg.py
ArianaGu: Yes it should work just fine, the sweep never reads pixel sizes, the transfer function is recomputed at the crop's shape from the config's own `yx_pixel_size`, and the Otsu window is a 3×3 pixel box.

waveorder/optim/autoreg.py
ArianaGu: Agreed, and thanks for offering to take it in the follow-up. Once the shared function exists, `_inverse_filter` here can go.

docs/examples/cli/configs/phase_3d_autoreg.yml
ArianaGu: I was thinking it would be good to keep the original behavior untouched by default, since this is a new feature, having it explicit in the config name made it clear which run was testing it. But yes we can consolidate as you suggested: `phase_3d.yml` now carries the `auto_regularization` block (model defaults, `report_path: null`) and the separate `_autoreg` config is gone.

docs/examples/cli/phase_3d_autoreg.sh
ArianaGu: see the reply to the last comment

waveorder/api/_settings.py
ArianaGu: fixed in `tab_recon.py`. An `Optional[Model]` sub-container now starts with an `enable` checkbox, ticked only when the loaded config carries the block, and `get_pydantic_kwargs` submits `None` when it's off. Also fixed `Optional[str]` leaves (`report_path`, `device`) coming back as the string `"None"`, which would have written the report to a file called `None`.

waveorder/optim/autoreg.py
ArianaGu: you're right that they're always `np.ndarray`, so it's `.tolist()` directly now. `warnings` stays `list(...)` since it's a plain `list[str]`.

waveorder/optim/autoreg.py
ArianaGu: You're right that message would have been swallowed by `wo rec`. Now generalized it: `AutoRegularizationWarning(UserWarning)` is now the family, `AutoRegularizationIgnoredWarning` subclasses it, the `filterwarnings("default", …)` carve-out sits on the base, and `warn_all` emits the base class; the existing edge-of-sweep test now asserts that category.

waveorder/optim/autoreg.py
ArianaGu: You're right, and it'a a clean change. It didn't matter too much for me because the example datasets I was given have 12-16 slices, but it matters a lot for the 2000+ slices case you mentioned. `select_regularization` now reconstructs one strength at a time and keeps only what the rule needs: the two norms for `l_curve`, and per-slice Otsu CNR plus per-slice local-variance spread for `otsu_cnr`, with the shared z slice chosen afterwards from the spread averaged over the sweep. Same slice, same numbers as before, without holding the volumes; `otsu_cnr_scores(recons)` keeps its batched signature so the reference fixture still pins the numbers.


Measured on a synthetic 256×256 crop with a real fluorescence OTF, 25 strengths. At Z = 400, peak memory during the sweep drops from 6 GiB to 1 GiB for `otsu_cnr` and from 14 GiB to 0.4 GiB for `l_curve`, with identical picks and bit-identical `otsu_cnr` scores; wall time unchanged. At the Z = 2000 you mentioned, the new path peaks at 7.8 GiB (`otsu_cnr`) and 4.4 GiB (`l_curve`) — against the ~13 GiB the old path needed for the reconstruction stack alone.

The result now also records `scored_z_index` (the shared slice), echoed by the CLI and in the report.

tests/optim/test_autoreg.py
ArianaGu: Done: `test_select_regularization_scores_the_model_reconstructions` runs `select_regularization` with `apodization_rolloff=0.25` on an undersampled configuration, reconstructs at the swept strengths through the models' `apply_inverse_transfer_function` with the same rolloff, and requires the sweep's scores (or norms) to match — and checks that the unapodized control differs, so the test isn't vacuous. Parametrized over both models; `l_curve` only without z padding, since its norms live on the padded grid the model API doesn't return. The `_sweep_one` tests stay as the finer-grained pointer to which piece drifted.

ArianaGu:
> @ArianaGu
> 
> > Note: wo rec cannot read big-endian (>u2) zarrs, as the zarrs codec pipeline raises "Non-native byte order not supported", with or without this change. Unrelated, but worth an issue.
> 
> Huh, how'd you run into this issue?

Talon gave me three example datasets to test on. One of them (the neuromast DIA volume) is OME-Zarr v2 stored big-endian, `dtype: ">u2"` (blosc/zstd); the two viral-sensor volumes are `<u2` and read fine. I don't know what wrote it. iohub 0.3.11 defaults to the `zarrs-python` implementation (`iohub/core/registry.py: _default = "zarrs-python"`), and reading any chunk raises

    RuntimeError: Non-native byte order not supported

Repro (zarrs 0.2.3, zarr 3.3.0):

    from iohub import open_ome_zarr
    ds = open_ome_zarr(".../000_2025-02-28-neuromast_DIA.zarr/0/0/0", layout="fov", mode="r")
    ds.data[0, 0, 0, :8, :8]  # RuntimeError

`open_ome_zarr(..., implementation="zarr-python")` reads it fine. `wo rec` uses the default, so that file can't go through the CLI at all; I've been reading it with zarr-python outside `wo`. It looks like an iohub matter (fall back to zarr-python, or byte-swap, on a non-native dtype) rather than a waveorder one

